Issue #25, opened 2022-07-16
Map setting widget - Requirements


nina992:
###  **Todo**

-  Create widget plug-in to show map setting and modify them
-  The map settings that will be displayed are : 

1.   Map view : 
      - [ ] Choose between ( 3D Terrain - 3D smooth - 2D)
      - [ ] Check box to define whether Terrain hides the underground features 
     
2.    Base Map 

3.    Image optimization
      - [ ] Checkbox to check if always show or not 
5.    Raster Map Quality
      - [ ] Range slider between quality and performance 

![image](https://user-images.githubusercontent.com/89770889/179356268-e7978de7-5234-4718-b607-268b7863fe6d.png)

 ### Design 
![image](https://user-images.githubusercontent.com/89770889/179688320-278bba48-662e-451a-b17e-190ce607736b.png)

lavalse:
Hey, @nina992 I changed a little bit of design for the map setting. 
![image](https://user-images.githubusercontent.com/3818046/179647194-7d8faece-1fc4-41a6-874a-072295795932.png)
We don't need to set the timeline plugin open or not here.

lavalse:
And also, one of the main feature we have here is the sider-bar will have a different UI based on whether it is inside the Re:Earth editor . Please check [here](https://www.figma.com/file/xWM8xrtIAh5oMj0seLd9Jd/view-2.0?node-id=534%3A63367).
About the detail of tech. I guess we still need to discuss more.

lavalse:
And also I m not sure why there is a **Enter the basement** checkbox in your screenshot. Because on my PC it looks like 
![image](https://user-images.githubusercontent.com/3818046/179647848-9129552c-d3b8-4324-b635-3d7adaa26f8c.png)
No that checkbox 😄. What happened ...

airslice:
@lavalse If you use chome `translate to English` then it will be translated as `enter the basement`.

nina992:
@lavalse  yes if I translate the page it shows this enter basement check box  , what is the right translation

lavalse:
@nina992 okok, I see. I also don't what that check does. Let's me check a little bit. 😄

lavalse:
OH. I see. It allows users to zoom in and can get into the earth. Then the land will become the sky. So funny. I think we don't need to support now. So ignore it now, please

PR #28, opened 2022-08-09
feat: create Map setting plugin for plateau view

nina992:
### Overview
The main purpose is to create plugins that shows plateau view functionalities, and one of them map setting 
please refer to this https://github.com/eukarya-inc/reearth-plateauview/issues/25
  
### What I've done
1. work on plateau view frontend structure (components and styles )
2. create the map setting plugin 
### What I haven't done
1. internationalization
2. show title above the tiles (base map images) 
### Screenshot


plugin/package.json
```diff
@@ -36,23 +36,30 @@
...
     "vite": "^3.0.4",
     "vite-plugin-cdn-import": "^0.3.5",
     "vite-plugin-dts": "^1.4.1",
     "vite-plugin-externals": "^0.5.1",
     "vite-plugin-singlefile": "^0.11.0",
+    "vite-plugin-svgr": "^2.2.1",
     "vitest": "^0.20.3"
   },
   "dependencies": {
+    "eslint-plugin-css-modules": "^2.11.0",
```
KaWaite, line 62: should be a devDep

plugin/vite.config.datacatalog-web.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="vitest" />
 /// <reference types="vite/client" />
+/// <reference types="vite-plugin-svgr/client" />
```
KaWaite, line 3: I don't think you need this (or the other references) inside the config files since you have the plugin/env.d.ts file with them

plugin/web/components/Wrapper.less
```diff
@@ -0,0 +1,11 @@
+.wrapper {
```
KaWaite, line 1: This Wrapper.less file should be with the rest of the less files, right?

plugin/web/components/Wrapper.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import "./Wrapper.less";
+
+
```
KaWaite, line 4: Seems like eslint isn't setup properly to fix on save.

plugin/web/components/layout/WdContent.tsx
```diff
@@ -0,0 +1,45 @@
+import { Content } from "antd/lib/layout/layout";
+import React,{ memo, ReactNode } from "react";
+
+import "../../../node_modules/antd/dist/antd.less";
```
KaWaite, line 4: Please avoid urls like this.
Also, I think except for really special cases, most things in node_modules are accessible through the dep you installed. So this less file should be retrievable by `antd/dist/antd.less`

plugin/web/components/layout/WdContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const WdContent: React.FC<Props> = ({ className, current, children }) => {
+  return (
+    <Content className={className}>
+      {
+        {
+          mapSetting: <MapSettingTab />,
+          shareNprint: <ShareTab />,
+          about: <InfoTab />,
+        }[current]
```
KaWaite, line 39: Can you explain this part of code? I have never seen something like this before and checking the antd documentation I can see it either

plugin/web/components/layout/WdContent.tsx
KaWaite: It seems you are doing this in multiple places. But can't you import this less file into your top level component (probably main.tsx) and it works for all? I don't think you need to import it anywhere else after that.

plugin/web/components/layout/WdHeader.tsx
```diff
@@ -0,0 +1,38 @@
+import { CloseOutlined } from "@ant-design/icons";
+import { Button, Menu, MenuProps, Row } from "antd";
+import { Header } from "antd/lib/layout/layout";
+import { MenuInfo } from "rc-menu/lib/interface";
```
KaWaite, line 4: Why do you need rc-menu's MenuInfo type? Doesn't seem like you are using rc-menu for anything else

plugin/web/components/layout/WdLayout.tsx
```diff
@@ -0,0 +1,65 @@
+import { Layout, MenuProps } from "antd";
+import { MenuInfo } from "rc-menu/lib/interface";
+import React,{ memo, useMemo, useState } from "react";
+
+import "../../../node_modules/antd/dist/antd.less";
+import "../../styles/style.less";
```
KaWaite, line 6: Same with this file as I said in another comment: I think you just need to import this in the main.tsx file and then good to go. Don't need to import in many components

plugin/web/components/layout/WdLayout.tsx
```diff
@@ -0,0 +1,65 @@
...
+    icon: <Info />,
+  },
+  {
+    key: "template",
+    icon: <Template />,
+  },
+];
+export type Props = {
+  className?: string;
+  isInsideEditor: boolean;
```
KaWaite, line 42: What is this isInsideEditor doing? Why does it slice up the items when in the editor and doesn't outside?

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
KaWaite, line 1: Think this file isn't needed

plugin/web/components/layout/WdContent.tsx
nina992: this something like inline enum , you can use it instead of if , else conditions or switch

plugin/web/components/layout/WdLayout.tsx
nina992: I added this attribute so when we start the implementation we can pass it to define if we are inside reearth editor or out of it , so the tabs that appears in the side bar will be different

PR #44, opened 2022-11-10
refactor: move the utils to global scope

nina992:
## Overview
This pr is to move the utils (general function that are used for general purposes like postMsg to reearth plugin Api)


plugin/web/types.ts
```diff
@@ -0,0 +1,9 @@
+type actionType =
+  | "updateOverrides"
+  | "screenshot"
+  | "screenshot-save"
+  | "modal-open"
+  | "modal-close"
+  | "msgFromModal"
+  | "minimize";
```
KaWaite, line 8: Sorry can you keep the `actionType` extension specific? To do that, you'll have to make PostMessageProps generic, so something like:
```
export type PostMessageProps<T> = { action: T; payload?: any };
```

PR #49, opened 2022-11-22
feat(server): 3D Tiles indexer

pyshx:
### To produce indexes one needs the following files:
1) Path to tileset.json file and associated b3dm files
2) An index configuration file (See [sample configuration](https://github.com/eukarya-inc/reearth-plateauview/blob/feat/indexer/server/indexer/testdata/config.json)).
3) An output directory to write the index

This generates an index in the given output directory. Check the indexRoot.json to see how the index is structured.

### indexRoot.json

`indexRoot` captures the overall structure of the index. It has the following fields:

- `idProperty`
  - Required
  - Name of the feature property that is used as ID for indexing. This is also
    sometimes used by the catalog item to uniquely identify & highlight the
    selected feature.
- `resultsDataUrl: string`
  - Required
  - URL of the CSV results data file mapping a feature by
    its ID to result data associated with the feature.
- `indexes: map[string]Index`
  - Required
  - An object whose keys are the property names and values are the
    corresponding Index definition.

### Results data file

A CSV file mapping an indexed feature to result data using its ID.

eg:

```
"_gml_id","latitude","longitude","height"
"abc1","45.0","100.0","120"
```

It should contain a header for each column. It should also have a column for the `idProperty` specified in the `indexRoot.json` file.

- `latitude`
  - Required
  - The latitude of the feature
- `longitude`
  - Required
  - The longitude of the feature
- `height`
  - Optional
  - The height of the feature

### Index types
**Note: Currently, we only support `enum` type of indexes.**

#### Enum Index

The enum index is useful for searching the fixed list of strings, eg: Roof material property can have a fixed set of values like Tile, Metal, Fiberglass, Concrete, Plastic, etc. An enum index contains a sub-index for each of its values. The sub-index is simply a list of feature IDs that have that value.

##### Definition

- `Kind: "enum"`
  - Required
- `Values: map[string]EnumValue`
  - Required
  - An object whose keys are the enum string and value defines the enum value index.

eg:

```
"建物利用現況_小分類": {
      "Kind": "enum",
      "Values": {
        "サービス工業施設（Ａ）（自動車修理工場）": {
          "Count": 404,
          "Url": "4-11.csv",
        },
        "サービス工業施設（Ｂ）（原動力を使用する床面積が５０㎡を超える工場、印刷、木材、石材の引割、機械撚糸、魚肉の練製品、セメント製品の製造、メッキ、合成樹脂の射出成形）": {
          "Count": 1421,
          "Url": "4-17.csv",
        },
        "作業所併用住宅（住宅と工業系用途を併用している場合）": {
          "Count": 1926,
          "Url": "4-15.csv",
        },
 }
```

##### Enum value index

Defines the index for a single enum member.

###### Definition

- `Count: int`
  - Required
  - Number of features that have this enum value.
- `Url: string`
  - Required
  - URL of the enum value index file.

###### Enum value index file

The enum value index file is a CSV file with a single column named `dataRowId` which is the index of the corresponding feature in the results data file.

eg:

```
"dataRowId"
"1"
"2"
...
"546205"
```


server/indexer/index.go
```diff
@@ -0,0 +1,17 @@
...
+
+type IndexRoot struct {
+	ResultDataUrl string
+	IdProperty    string
+	Indexes       map[string]interface{}
+}
+
+type EnumIndex struct {
+	Kind   string
+	Values map[string]*EnumValue
```
rot1024, line 11:
```suggestion
	Kind   string                `json:"kind"`
	Values map[string]*EnumValue `json:"values"`
```

All struct fields that can be serialized or deserialized as JSON should have a `json` tag. CamelCase with the first letter capitalized is not very common in JSON.

server/indexer/testdata/config.json
```diff
@@ -0,0 +1,21 @@
...
+      "kind": "enum"
+    },
+    "建物利用現況_中分類": {
+      "kind": "enum"
+    },
+    "建物利用現況_小分類": {
+      "kind": "enum"
+    }
+  }
+}
```
rot1024, line 21: Remove this file

server/indexer/util_test.go
```diff
@@ -0,0 +1 @@
+package indexer
```
rot1024, line 1: Remove this file or add unit tests

server/indexer/cesium_math.go
```diff
@@ -0,0 +1,43 @@
...
+const (
+	_EPSILON1           = 0.1
+	_EPSILON12          = 0.000000000001
+	_EPSILON14          = 0.00000000000001
+)
+
+const (
+	_PI     = math.Pi
+	_TWO_PI = 2 * math.Pi
+	_DEGREES_PER_RADIAN = 180.0 / math.Pi
```
rot1024, line 16:
```suggestion
	Epsilon1           = 0.1
	Epsilon12          = 0.000000000001
	Epsilon14          = 0.00000000000001
	Pi     = math.Pi
	TwoPi = 2 * math.Pi
	DegreesPerRadian = 180.0 / math.Pi
```

(Note that this suggestion is handwritten and not formatted)

Adding an underscore at the beginning is not particularly meaningful in Golang and is not very common. Golang rarely uses `snake_case`. Use `camelCase` (private) or `CamelCase` (public). In this case CamelCase is ok.

server/indexer/indexer_test.go
```diff
@@ -0,0 +1,15 @@
...
+var (
+	_TEST_TILESET_JSON = "testdata/tileset.json"
+	_TEST_CONFIG_JSON = "testdata/config.json"
+	_TEST_RESUILT_PATH = "testdata/result"
+)
+
+func TestIndexer(t *testing.T) {
+	if err := Indexer(_TEST_TILESET_JSON, _TEST_CONFIG_JSON, _TEST_RESUILT_PATH); err != nil {
+		t.Errorf("failed to generate indexes: %v", err)
+	}
```
rot1024, line 14: Do not run the test data file if it does not exist.

server/indexer/util.go
```diff
@@ -0,0 +1,128 @@
...
+	result := make([]float64, len(elem))
+	var err error
+	for i, v := range elem {
+		result[i], err = f(v)
+		if err != nil {
+			return nil, fmt.Errorf("failed to apply function the function: %w", err)
+		}
+	}
+	return result, err
+}
```
rot1024, line 109: You can use `util.TryMap` in `reearth/reearthx` instead of this function. It has `TryMap[T, K any]([]T, func(T) (K, error)) ([]K, error)`

server/indexer/indexer.go
```diff
@@ -0,0 +1,301 @@
...
+			return nil, fmt.Errorf("something went wrong at iterTile: %v", err)
+		}
+	}
+
+	return uniqueFeatures, nil
+
+}
+
+func Indexer(tilesetFile, indexConfigFile, outDir string) error {
+	tileset, err := tiles.Open(tilesetFile)
```
rot1024, line 229:
```suggestion
	tileset, err := tiles.Open("tileset.json") // or config.TilesetJSONPath
```

I believe the specification specifies that the file name is `tileset.json`, but if you need flexibility, you should be able to set that in `IndexerConfig`.

server/indexer/indexer_config.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+	"io"
+	"os"
+)
+
+type Index struct {
+	Kind string `json:"kind"`
+}
+
+type IndexesConfig struct {
```
rot1024, line 14: Better naming conventions:

```suggestion
type Config struct {
```

`Config` is enough, as users of this package requires writing `indexer.Config{}`, but it is easily understandable what config it is.

In general, if the package name is `indexer`:

```go
package indexer

type Indexer struct {} // it's ok
type IndexerHogehoge struct {} // -> Hogehoge

func NewIndexer() *Indexer // -> New
func NewIndexerHogehoge() *IndexerHogehoge // -> NewHogehoge
```

server/.gitignore
```diff
@@ -0,0 +1,3 @@
+indexer/testdata/data
+indexer/testdata/result
+indexer/testdata/tileset.json
```
rot1024, line 3:
```suggestion
indexer/testdata
```

rot1024 (changes requested):

server/indexer/indexer.go
```diff
@@ -0,0 +1,301 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("something went wrong at iterTile: %v", err)
+		}
+	}
+
+	return uniqueFeatures, nil
+
+}
+
+func Indexer(tilesetFile, indexConfigFile, outDir string) error {
```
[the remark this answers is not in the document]
pyshx: Gotcha!
In that case, it'll not just be IndexRoot which we'll have to return it be more like
```suggestion
func Indexer(config Config, fs io.FS) (IndexBuilder, []map[string]string, IndexRoot )
```


Context: 

- IndexBuilder stores the location of each feature at `resultsData.csv` which were getting saved as csv as well e.g `4-34.csv` in the image attached below
Capture

-  []map[string]string will the be the `resultsData.csv` content

server/indexer/index_builder.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	defer f.Close()
+
+	w := csv.NewWriter(f)
+	defer w.Flush()
+
+	var keys []string
+	for k := range data[0] {
+		keys = append(keys, k)
+	}
```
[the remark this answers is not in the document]
pyshx: Thanks.

rot1024 (approved):

PR #56, opened 2022-12-13
feat(plugin): add infobox to geolocation plugin created layer 

nina992:
# Overview
This is a new feature for the geoLocation widget - when getting current location we add Layer as marker to where this location is , in this pr we add infoBox to this Layer 
## Screenshot
image


plugin/src/geolocation.ts
```diff
@@ -24,6 +24,35 @@ reearth.on("message", ({ action, payload }: PostMessageProps) => {
...
+                  item_datatype: "number",
+                  item_datanum: payload.currentLocation.latitude,
+                },
+                {
+                  item_title: "Lng",
+                  item_datatype: "number",
+                  item_datanum: payload.currentLocation.longitude,
+                },
+              ],
+              paddingLeft: 24,
```
KaWaite, line 45: This `paddingLeft` isn't correct, right?

KaWaite (approved):

PR #58, opened 2022-12-20
feat(plugin): create help page for plateau view sidebar

nina992:
# Overview
This pr is  for the help page 
Help page shows multi selection to describe how system functionalities work.
## What I've done
- create the help page selection 
- create pop up for each embedded help page 
- create modals to show YouTube videos description 

## Screenshot
image
image

## Memo
Design Link : https://www.figma.com/file/xWM8xrtIAh5oMj0seLd9Jd/view-2.0?node-id=3233%3A86856&t=K3Bv1dgU866wDPRZ-0


plugin/vite.config.template.ts
```diff
@@ -31,7 +31,7 @@ export const web =
   }: {
     name: string;
     parent?: string;
-    type?: "modal" | "plugin" | "core";
+    type?: "modal" | "plugin" | "core" | "popup";
```
KaWaite, line 34: I think I made a mistake with "plugin", so if you can remove that that would be great!

nina992: you mean I can build the code without popup type , but what should be the type of popup page

nina992: image
image

KaWaite (approved):

PR #69, opened 2023-01-18
feat(plugin): Welcome Screen modal

nina992:
# Overview
Welcome Page for plateauView 
## Screenshot
image
image

## Memo
Design Link :https://www.figma.com/file/xWM8xrtIAh5oMj0seLd9Jd/view-2.0?node-id=3164%3A88520&t=LlqyXZ7LLIoEW9gT-0


plugin/src/sidebar.ts
```diff
@@ -14,11 +14,12 @@ let addedDatasets: string | undefined = undefined;
 
 let rawCatalog: CatalogRawItem[] = [];
 
-const doNotShowWelcome = true; // Make it `let doNotShowWelcome: boolean = false`, and then modify based on storage value when Storage API available
+let doNotShowWelcome = false; // Make it `let doNotShowWelcome: boolean = false`, and then modify based on storage value when Storage API available
 
 // let isMobile: boolean;
 
 reearth.ui.show(html, { extended: true });
+reearth.modal.show(welcomeScreenHtml, { width: 1811, height: 959 });
```
KaWaite, line 22: If you look at the design, you'll want to set the width and height based on the viewport (the X button is in the top right corner). 
Also, you will only want to show this modal IF there is doNotShowWelcome value in the storage. Having it here means it'll show every time the page is loaded. It also seems like there is no logic to check if there is a doNotShowWelcome value in storage, and if there is, and it is true, to NOT show the modal.

plugin/src/sidebar.ts
```diff
@@ -35,9 +36,15 @@ reearth.on("message", ({ action, payload }: PostMessageProps) => {
       },
     });
     reearth.clientStorage.setAsync("overrides", payload);
+    reearth.clientStorage.getAsync("doNotShowWelcome").then((value: any) => {
+      doNotShowWelcome = value;
+    });
```
KaWaite, line 41: The `getAsync` check is good here on `initSidebar`, but inside the `then()` you should be doing a bit more.
Here you should check if doNotShowWelcome is truthy, and if so, do nothing. If it is falsey (undefined or false), this is where you should do the `reearth.modal.show(welcomeScreenHtml, ....)`

KaWaite: Also, I don't think you'll need the `let doNotShowWelcome = false;` on line 17 anymore since checking the storage is enough

plugin/web/extensions/sidebar/modals/welcomescreen/component/Backdrop/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+type Props = {
+  onClose: () => void;
+};
+
+const Backdrop: React.FC<Props> = ({ onClose }) => {
+  return <BackdropWrapper onClick={onClose} />;
+};
+export default Backdrop;
+const BackdropWrapper = styled.div`
```
KaWaite, line 11:
```suggestion
const Backdrop: React.FC<Props> = ({ onClose }) => {
  return <BackdropWrapper onClick={onClose} />;
};

export default Backdrop;

const BackdropWrapper = styled.div`
```

plugin/web/extensions/sidebar/modals/welcomescreen/component/Backdrop/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+  position: fixed;
+  top: 0;
+  right: 0;
+  left: 0;
+  width: 100%;
+  height: 100vh;
```
KaWaite, line 17: height: 100%;

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
```diff
@@ -0,0 +1,191 @@
+import welcomeScreenVideo from "@web/extensions/sidebar/core/assets/welcomeScreenVideo.png";
+// import Backdrop from "@web/extensions/sidebar/modals/welcomescreen/component/Backdrop";
```
KaWaite, line 2: Should be using this Backdrop to wrap everything, right?

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  flex-direction: row;
+  align-items: flex-end;
+  padding: 0px;
+  gap: 119px;
+  position: absolute;
+  width: 742px;
+  height: 316px;
+  left: calc(50% - 742px / 2 + 0.5px);
+  top: calc(50% - 316px / 2 - 30.5px);
+`;
```
KaWaite, line 89: These styles need to be updated.
You set the width and height in the show API options so Wrapper needs to be 100% width and height, and then here would be where you set the background color, etc. Something like:
```
const Wrapper = styled.div`
   height: 100%;
   width: 100%;
   display: flex;
   justify-content: center;
  align-items: center;
  background: whateverTheSemiTransparentColorIs;
...
...
`;
```
The above will give you the full screen overlay and position the InnerWrapper (where the two buttons, the video and the checkbox will be) in the center of the screen.
With that, you can simplify InnerWrapper's CSS as you don't need to position absolute it, give the gap, etc

KaWaite: Also, it seems like a lot of the styled components' CSS is written assuming they need to position absolutely, which with the above changes you shouldn't need to do so much work.

plugin/web/extensions/sidebar/modals/welcomescreen/hooks.ts
```diff
@@ -0,0 +1,31 @@
...
+export default () => {
+  const [ShowVideo, setShowVideo] = useState(false);
+  const [dontShowAgain, setDontShowAgain] = useState(false);
+
+  const handleDontShowAgain = useCallback(() => {
+    setDontShowAgain(!dontShowAgain);
+    postMsg({
+      action: "storageSaveWelcomeScreen",
+      payload: { key: "doNotShowWelcome", value: dontShowAgain },
+    });
```
KaWaite, line 13: I feel like this postMsg should happen in the handleClose, and be called ONLY IF dontShowAgain is true. We don't need to save a falsey value.

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+const WelcomeScreen: React.FC = () => {
+  const { ShowVideo, handleDontShowAgain, dontShowAgain, handleShowVideo, handleClose } =
+    useHooks();
+
+  return (
+    <Wrapper>
+      <CloseButton>
+        <Icon size={32} icon="close" onClick={handleClose} />
+      </CloseButton>
+      {!ShowVideo ? (
```
KaWaite, line 17: I think the InnerWrapper should always be shown. 
Instead, can you make the VideoWrapper `display: none;` if showVideo is false. VideoWrapper will need to be positioned absolutely so it displays overtop the InnerWrapper content.

plugin/web/extensions/sidebar/modals/welcomescreen/component/Backdrop/index.tsx
nina992: removed - based on discussion with Red

plugin/web/extensions/sidebar/modals/welcomescreen/component/Backdrop/index.tsx
nina992: removed - based on discussion with Red

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: removed - based on discussion with Red

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: Done

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: based on discussion with red , the close button on video will return the user to the main welcome page so, I keep it as it it and handling close in video mode to return the user to the welcome page

plugin/web/extensions/sidebar/modals/welcomescreen/component/Backdrop/index.tsx
nina992: e8d02a5faeea4dc4732056ccaf88ca6f433e776c

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: e8d02a5faeea4dc4732056ccaf88ca6f433e776c

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: e8d02a5faeea4dc4732056ccaf88ca6f433e776c

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: e8d02a5faeea4dc4732056ccaf88ca6f433e776c

plugin/src/sidebar.ts
nina992: sorry it is for test purpose , removed

plugin/src/sidebar.ts
nina992: done

plugin/web/extensions/sidebar/modals/welcomescreen/component/Backdrop/index.tsx
nina992: e8d02a5faeea4dc4732056ccaf88ca6f433e776c

plugin/web/extensions/sidebar/modals/welcomescreen/component/Backdrop/index.tsx
nina992: e8d02a5faeea4dc4732056ccaf88ca6f433e776c

plugin/web/extensions/sidebar/types.ts
```diff
@@ -21,7 +21,8 @@ type ActionType =
   | "popupClose"
   | "msgToPopup"
   | "msgFromPopup"
-  | "minimize";
+  | "minimize"
+  | "storageSaveWelcomeScreen";
```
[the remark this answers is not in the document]
nina992: done

plugin/src/sidebar.ts
nina992: 2e7e5ada980318f9de64a960bc7f473718816a86

plugin/src/sidebar.ts
nina992: 2e7e5ada980318f9de64a960bc7f473718816a86

plugin/web/extensions/sidebar/types.ts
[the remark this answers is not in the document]
nina992: 2e7e5ada980318f9de64a960bc7f473718816a86

plugin/web/extensions/sidebar/modals/welcomescreen/hooks.ts
nina992: da1bbcf40d847aea1dc797d4d6b3493eaf5889a3

plugin/src/sidebar.ts
nina992: for viewport 3316a3a6b9f898adf24261ce66d0e2f4b61d8fcc

plugin/src/sidebar.ts
```diff
@@ -114,9 +120,15 @@ reearth.on("update", () => {
...
+  // if (e.isMobile !== isMobile) {
+  //   reearth.ui.postMessage({type: ""});
+  //   isMobile = e.isMobile;
+  // }
+
+  if (welcomPagIsOpen)
+    reearth.modal.update(welcomeScreenHtml, {
+      width: reearth.viewport.width,
+      height: reearth.viewport.height,
+    });
```
KaWaite, line 133: Please recheck the API. I believe `reearth.modal.update` only takes an object as a param with width, height and background fields.

plugin/src/sidebar.ts
```diff
@@ -35,9 +33,17 @@ reearth.on("message", ({ action, payload }: PostMessageProps) => {
...
+      if (!value && !reearth.scene.inEditor) {
+        reearth.modal.show(welcomeScreenHtml, {
+          width: reearth.viewport.width,
+          height: reearth.viewport.height,
+        });
+        welcomPagIsOpen = true;
+      }
+    });
+  } else if (action === "storageSaveWelcomeScreen") {
+    reearth.clientStorage.setAsync(payload.key, payload.value);
```
KaWaite, line 46: This can just use the `storageSave` action

plugin/src/sidebar.ts
```diff
@@ -14,10 +14,8 @@ let addedDatasets: string | undefined = undefined;
 
 let rawCatalog: CatalogRawItem[] = [];
 
-const doNotShowWelcome = true; // Make it `let doNotShowWelcome: boolean = false`, and then modify based on storage value when Storage API available
-
 // let isMobile: boolean;
-
+let welcomPagIsOpen = false;
```
KaWaite, line 18: welcomPagIsOpen -> welcomePageIsOpen

nina992: done

plugin/src/sidebar.ts
nina992: done

plugin/src/sidebar.ts
nina992: done

plugin/src/sidebar.ts
nina992: 6dc6f5efc904b2988a326ea6674b939ecdca1183

plugin/src/sidebar.ts
nina992: 6dc6f5efc904b2988a326ea6674b939ecdca1183

plugin/src/sidebar.ts
nina992: 6dc6f5efc904b2988a326ea6674b939ecdca1183

plugin/src/sidebar.ts
```diff
@@ -14,10 +14,8 @@ let addedDatasets: string | undefined = undefined;
 
 let rawCatalog: CatalogRawItem[] = [];
 
-const doNotShowWelcome = true; // Make it `let doNotShowWelcome: boolean = false`, and then modify based on storage value when Storage API available
-
 // let isMobile: boolean;
-
+let welcomePagIsOpen = false;
```
KaWaite, line 18: welcomePageIsOpen
Missing the `e` on Page

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+const InnerWrapper = styled.div`
+  display: flex;
+  flex-direction: row;
+  align-items: flex-end;
+  padding: 0px;
+  gap: 119px;
+  width: 742px;
+  height: 316px;
+  left: calc(50% - 742px / 2 + 0.5px);
+  top: calc(50% - 316px / 2 - 30.5px);
```
KaWaite, line 106: I don't think this is doing anything since InnerWrapper isn't being positioned absolutely.

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+  color: #fafafa;
+`;
+
+const TryMapWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  justify-content: flex-end;
+  padding: 0px;
+  gap: 55px;
```
KaWaite, line 127: I think you are using gap here, in BtnsWrapper, etc almost like padding, no? Gap should be used to create a gap between the children inside a flex container. If I am right, please switch from gap to padding in each of these components instead.

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+  height: 46px;
+  left: calc(50% - 178px / 2 + 0.5px);
+  top: calc(50% - 22px / 2 + 178.5px);
+`;
+
+const VideoWrapper = styled.div`
+  width: 1142px;
+  height: 543px;
+  left: calc(50% - 1142px / 2 + 0.5px);
+  top: calc(50% - 543px / 2);
```
KaWaite, line 207: I think this is same as above and not really doing anything

plugin/src/sidebar.ts
nina992: done

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: it is to set the gap between the inner wrapper and the checkbox wrapper

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: I will remove it and test it

nina992: done

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: done

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: I replaced it with padding top for the checkbox wrapper

plugin/src/sidebar.ts
nina992: a36d7887f9fae43211d25e5df9545ce613bbccec

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: a36d7887f9fae43211d25e5df9545ce613bbccec

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: a36d7887f9fae43211d25e5df9545ce613bbccec

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: a36d7887f9fae43211d25e5df9545ce613bbccec

KaWaite (approved):

PR #74, opened 2023-01-25
feat(plugin): data catalog your data

nourbalaha:
In this PR I did the following:
- I added styles to the my data tab in data catalog
- I added functionality to my data tab for both local data and web data in data catalog


plugin/web/extensions/sidebar/modals/datacatalog/components/content/YourDataPage/LocalDataTab.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const beforeUpload = useCallback(
+    (file: RcFile, files: RcFile[]) => {
+      // Catalog Item
+      const filename = file.name;
+      const id = "id" + Math.random().toString(16).slice(2);
+      const url = URL.createObjectURL(file);
+      const item = {
+        id: id,
+        description:
+          "This file only exists in your browser. To share it, you must load it onto a public web server.",
```
KaWaite, line 31: Can you find a way to check plateau v1's Japanese for this line? Or ask a Japanese member to translate?

plugin/web/extensions/sidebar/modals/datacatalog/components/content/YourDataPage/LocalDataTab.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import FileTypeSelect from "./FileTypeSelect";
+
+type Props = {
+  onOpenDetails?: (data?: CatalogItem) => void;
+};
+
+const LocalDataTab: React.FC<Props> = ({ onOpenDetails }) => {
+  const [fileList, setFileList] = useState<UploadFile[]>([]);
+  const fileFormats = ".kml,.czml,.topojson,.geojson,.json,.gltf,.glb";
```
KaWaite, line 16: You can move this outside of the LocalDataTab component since it'll never change

plugin/web/extensions/sidebar/core/processCatalog.ts
```diff
@@ -5,8 +5,8 @@ type ModelType = "plateau" | "usecase" | "dataset";
 export type CatalogRawItem = {
   id: string;
   name?: string;
-  prefecture: string; // 都道府県
-  cityName: string; // city_name 市区町村名
+  prefecture?: string; // 都道府県
+  cityName?: string; // city_name 市区町村名
```
KaWaite, line 9: Hmmm...when I think about it, leaving CatalogRawItem as it was, and then creating a UserData (or better name) type specific to user data is better.
The reason is that the user uploaded data will never be a part of the PLATEAUデータセット "Catalog". 
Hmmm, instead, I think you can ignore (for now) converting types.

plugin/web/extensions/sidebar/modals/datacatalog/components/content/YourDataPage/LocalDataTab.tsx
nourbalaha: Done!

plugin/web/extensions/sidebar/core/processCatalog.ts
nourbalaha: OK! So I think that I will create the following type instead.
```ts
type UserDataItem = {
  id: string;
  name?: string;
  description?: string;
  data?: string;
  dataUrl?: string;
}
```
What do you think?

plugin/web/extensions/sidebar/modals/datacatalog/components/content/YourDataPage/LocalDataTab.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { useCallback, useMemo, useState } from "react";
+
+import FileTypeSelect, { FileType } from "./FileTypeSelect";
+
+type Props = {
+  onOpenDetails?: (data?: UserDataItem) => void;
+  setSelectedLocalItem?: (data?: UserDataItem) => void;
+};
+
+const fileFormats = ".kml,.kmz,.csv,.czml,.gpx,.geojson,.json,.zip";
```
KaWaite, line 16: Can you move these in to the FileTypeSelect and then import here from that file? That way all file type information is in one spot

KaWaite (approved):

plugin/web/extensions/sidebar/modals/datacatalog/components/content/YourDataPage/LocalDataTab.tsx
nourbalaha: Done!

PR #75, opened 2023-01-25
feat(plugin): add infobox UI

airslice:
# Overview

This PR adds infobox UI.

_Data/communication part is NOT included in this PR_

Some (DRAFT) expect for reearth & plateauview backend:
- Reearth: Get selected primitive(s) info with plugin API.
- Reearth: Disable built-in property infobox override for 3d tiles primitives.
- Plateauview backend: Store all 3d tiles <-> type data.
- Plateauview sidebar: Store added 3d tiles <-> type <-> layerId data.

Not sure select is by layer or by primitive in new layer system.
- Not sure if the infobox will close/open (rerender for the plugin block) when select another primitive.

## What I've done

- UI of infobox block.

## How I tested

UI preview project [here](https://test.reearth.dev/edit/01gqm4rva8hnj3y0am1qbr18t4).


plugin/web/extensions/infobox/core/utils/index.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function postMsg(action: ActionType, payload?: any) {
+  if (parent === window) return;
+  parent.postMessage({
+    action,
+    payload,
+  });
+}
+
+export function generateId() {
```
KaWaite, line 11: is this being used anywhere?

plugin/web/extensions/infobox/core/index.html
```diff
@@ -0,0 +1,20 @@
...
+<link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+<link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
+<title></title>
+<style>
+  html, body{
+    background: transparent;
+    width: 100%;
+    height: 100%;
+    min-height: 500px;
```
KaWaite, line 12: Is this going to break in mobile?

plugin/web/extensions/infobox/core/components/viewer/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  publicSettings: PublicSetting[];
+};
+
+type PropertyItemType = {
+  key: string;
+  title: string;
+  value?: any;
+};
+
+const Viewer: React.FC<Props> = ({ primitive, publicSettings, key, ...props }) => {
```
KaWaite, line 18: Are you expecting anything in the `...props`? Because you destructure (as far as I can tell) all your props.

plugin/web/extensions/infobox/core/components/viewer/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [propertyList, setPropertyList] = useState<PropertyItemType[]>([]);
+
+  useEffect(() => {
+    const propertyItem: PropertyItemType[] = [];
+    const publicSetting = publicSettings.find(ps => ps.type === primitive.type);
+    if (!publicSetting || publicSetting.properties.length === 0) {
+      primitive.properties.forEach(property => {
```
KaWaite, line 25: Typically you should avoid forEach because it'll mutate the original. Are you doing this on purpose, or perhaps you should make the primitive as a useState and use its setter function to modify it?
If you think this is fine, please let me know as I am a little unsure when mutating the original with a forEach is actually desirable (at least inside of React).

plugin/web/extensions/infobox/core/components/viewer/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+        propertyItem.push({
+          ...property,
+          title: property.key,
+        });
+      });
+    } else {
+      publicSetting.properties
+        .filter(psp => !psp.hidden)
```
KaWaite, line 33: Sorry I could be wrong, but won't this hide hidden (non-public) properties ALWAYS? Even if editor mode?

plugin/web/extensions/infobox/core/components/editor/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+import type { PropertyItem as PropertyItemType } from "./PropertyItem";
+import PropertyItem from "./PropertyItem";
+
+type Props = {
+  publicSetting: PublicSetting;
+  primitives: Primitive[];
+  savePublicSetting: (publicSetting: PublicSetting) => void;
+};
+
+const Editor: React.FC<Props> = ({ primitives, publicSetting, savePublicSetting, ...props }) => {
```
KaWaite, line 18: Same with this ..props here

plugin/web/extensions/infobox/core/components/editor/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const [propertyList, setPropertyList] = useState<PropertyItemType[]>([]);
+
+  useEffect(() => {
+    const proptyItems: PropertyItemType[] = [];
+    const primitive = primitives.findLast(p => p.type === publicSetting.type);
```
KaWaite, line 23: Any reason you are finding the last primitive that meets the requirement? I thought you and I discussed on Wednesday that it made sense to show the values for the currently selected model, no?

plugin/src/infobox.ts
```diff
@@ -0,0 +1,37 @@
...
+      payload: reearth.scene.inEditor,
+    });
+  } else if (action === "savePublicSetting") {
+    getSidebarId();
+    if (!sidebarId) return;
+    reearth.plugins.postMessage(sidebarId, {
+      action: "savePublicSetting",
+      payload,
+    } as SavePublicSetting);
+    console.log(payload);
```
KaWaite, line 35: There is some code here and there that is for devs, like this console log. If it is on purpose to help you out once the new layer system is merged and to finish the logic connecting the infobox to the new layer system, it is fine. I just want to make sure you are leaving it on purpose and not forgetting

plugin/web/extensions/infobox/core/utils/index.ts
airslice: Not in use actually. I leave it here for primitives, i'm not sure if primitive will get a unique id or not. But seems that it's also okey if no id, so emmm yes maybe better remove it.

plugin/web/extensions/infobox/core/index.html
airslice: I think it's fine since infobox has a built-in scroll wrapper outside.

plugin/web/extensions/infobox/core/components/viewer/index.tsx
airslice: This is from Antd Collapse component, it has some difficulty in wrap Panel as a child component and this is a solution for it. Please check [here](https://github.com/react-component/collapse/issues/73#issuecomment-878312513).

plugin/web/extensions/infobox/core/components/editor/index.tsx
airslice: Same as the above.

plugin/web/extensions/infobox/core/components/editor/index.tsx
airslice: Currently i assume that new layer system supports multiple select (both of layers and primitives). So consider with multiple primitive selected, here should show the last selected one's value. (And i assume the last selected will be the last in array).

plugin/src/infobox.ts
airslice: This is the only console.log left in code. Actually this part is already touching the communication part so it may update in the future. I leave it here for confirm the save button works well.

airslice: Removed.

plugin/web/extensions/infobox/core/components/editor/index.tsx
KaWaite: Ahhh okay, gotcha

KaWaite (approved):

PR #81, opened 2023-01-30
feat(plugin): add building search UI

airslice:
# Overview

This PR adds building search UI.

Building search is a popup of sidebar widget.

_Data/communication part is NOT included in this PR_

## What I've done

- UI of building search popup.

## How I tested

UI preview project [here](https://test.reearth.dev/edit/01gr10f0e34kha1f3hnsgxfcpn). 

This preview project will auto popup when initialize the sidebar.

## Memo


plugin/web/extensions/sidebar/popups/buildingSearch/components/hooks.ts
```diff
@@ -0,0 +1,214 @@
...
+    },
+    [handleResize],
+  );
+
+  useEffect(() => {
+    addEventListener("message", onMessage);
+    return () => {
+      removeEventListener("message", onMessage);
+    };
+  }, [onMessage]);
```
KaWaite, line 193: I think if you add anything to the deps array of a useEffect with addEventListener you will get multiple even listeners in the lifetime of the the component. Just move the onMessage function inside the useEffect.

KaWaite (approved):

plugin/src/sidebar.ts
```diff
@@ -55,6 +56,8 @@ const sidebarInstance: PluginExtensionInstance = reearth.plugins.instances.find(
 
 // ************************************************
 // initializations
+let buildingSearchIsOpen = false;
```
KaWaite, line 59: Can you move this up where the other xxIsOpen = false let statements?

plugin/web/extensions/sidebar/popups/buildingSearch/components/hooks.ts
airslice: Emmm i think even move it inside, the `handleResize` will also become the dependency. I think it's reasonalble to list as dependency and when dependency changes the returned clear function will remove previous event listner so there won't be multiple, right?

plugin/src/sidebar.ts
airslice: Sure. Updated.

PR #83, opened 2023-01-31
style(plugin): fix file-tree scroll

nourbalaha:
Done:
- Fix file-tree vertical scroll

### DIFF
--- a/plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
+++ b/plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
@@ -59,27 +59,34 @@ const FileTree: React.FC<Props> = ({ catalog, onOpenDetails }) => {
   }, []);
 
   return (
-    <Tree>
-      {catalog.map(item =>
-        TreeBuilder({
-          item,
-          selectedId,
-          nestLevel: 1,
-          onOpenDetails,
-          onSelect: handleSelect,
-        }),
-      )}
-    </Tree>
+    <TreeWrapper>
+      <Tree>
+        {catalog.map(item =>
+          TreeBuilder({
+            item,
+            selectedId,
+            nestLevel: 1,
+            onOpenDetails,
+            onSelect: handleSelect,
+          }),
+        )}
+      </Tree>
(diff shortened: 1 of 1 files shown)


plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
```diff
@@ -59,27 +59,34 @@ const FileTree: React.FC<Props> = ({ catalog, onOpenDetails }) => {
...
+          TreeBuilder({
+            item,
+            selectedId,
+            nestLevel: 1,
+            onOpenDetails,
+            onSelect: handleSelect,
+          }),
+        )}
+      </Tree>
+    </TreeWrapper>
```
KaWaite, line 74: Unless you plan on writing a new tree component for mobile, I don't think setting a 400px height on the TreeWrapper will be appropriate for the mobile version.

nourbalaha: I think that we can add a media query to adjust the height for the mobile version. What do you think??

KaWaite: Media queries in the typical sense don't work inside of plugin extensions. What you will need to use is the reearth.viewport.isMobile API. But, inside the Mobile sidebar we already know it is mobile, so perhaps just adding a isMobile prop to the FileTree is enough, and then you can customize the height based on that.
Before merging and doing the isMobile approach, can you test quickly if setting the height on the TreeWrapper to 100% works? Because I believe in mobile it'd need to be that anyways.

KaWaite (approved):

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
nourbalaha: I tried to add a 100% height and this is what I've got:
Screenshot 0005-02-01 at 11 21 29 AM

That is why I went with adding the `isMobile` prop approach instead.

PR #98, opened 2023-02-09
feat(plugin): data catalog search and filter

nourbalaha:
Done:
- Implement data catalog's tag filter.
- Implement data catalog's search.
- Add expand all to file tree.
- Fix mobile styles.
- Refactor.


plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/index.tsx
```diff
@@ -11,18 +11,29 @@ export type Props = {
...
+  const handleFilter = useCallback((filter: FilterType) => {
+    setFilter(filter);
+  }, []);
+
   const handleTagSelect = useCallback(
     (tag: Tag) =>
-      selectTags(tags => (tags.includes(tag) ? [...tags.filter(t => t !== tag)] : [...tags, tag])),
-    [],
+      selectTags(tags => {
+        const selected = tags.includes(tag) ? [...tags.filter(t => t !== tag)] : [...tags, tag];
```
KaWaite, line 32: I think you can just do `tags.filter(t => t !== tag)`. The way you have it now you are basically creating an array just to spread it in to another array.

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/index.tsx
```diff
@@ -43,24 +44,12 @@ function typeFilter(catalog: CatalogRawItem[]): DataCatalog {
...
-      type: "group",
-      name: p,
-      children: items,
-    };
-  });
-  return filteredCatalog;
+function tagFilter(catalog: CatalogRawItem[], tags?: Tag[]): DataCatalog {
+  return catalog
+    .filter(item =>
+      item.tags?.some(tag => tags?.some(selectedTag => tag.name === selectedTag.name)),
```
KaWaite, line 50: I could be wrong, but can you simplify this by doing:
```
return catalog
    .filter(item =>
      item.tags?.some(tag => tags?.includes(tag)),
```

KaWaite (approved):

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/index.tsx
nourbalaha: This might work if `tag` was of type string (strings are passed by value in javascript). 
However, since `tag` is an object (objects are passed by reference in javascript) the `.includes()` method would not work.

PR #111, opened 2023-02-15
feat(plugin): story telling behavior updates

airslice:
This is a small PR update some behaviors about story telling:
- Forbidden scene save with empty title.
- Create new scene after save scene info.
- Save story on every edit of scenes. Share only set sidebar to share page.

Tested with [this project.](https://test.reearth.dev/edit/01gpmnf4ah0syavpemp8ywhhsf)

### DIFF
--- a/plugin/web/extensions/sidebar/core/components/hooks.ts
+++ b/plugin/web/extensions/sidebar/core/components/hooks.ts
@@ -254,16 +254,22 @@ export default () => {
     [backendURL, backendAccessToken],
   );
 
-  const handleStoryShare = useCallback((story: Story) => {
+  const handleStoryShare = useCallback(() => {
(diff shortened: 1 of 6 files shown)


plugin/web/extensions/sidebar/core/components/hooks.ts
```diff
@@ -254,16 +254,22 @@ export default () => {
     [backendURL, backendAccessToken],
   );
 
-  const handleStoryShare = useCallback((story: Story) => {
+  const handleStoryShare = useCallback(() => {
+    setCurrentPage("share");
+  }, []);
```
KaWaite, line 259: This isn't really handling story sharing, so can you rename this to handleGoToShare or something?

KaWaite (approved):

PR #114, opened 2023-02-15
feat(plugin):create the field components style code and real time update 

nina992:
# Overview
- Create the field components style code and real time update.
- Add validation prefix to url in button link  
  
## Memo
Design Link : https://www.figma.com/file/xWM8xrtIAh5oMj0seLd9Jd/view-2.0?node-id=3009%3A88041&t=woyMZiyzUrtI5MSQ-0


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/types.ts
```diff
@@ -62,17 +65,18 @@ export type Legend = FieldBase<"legend"> & {
   items?: LegendItem[];
 };
 
-// type Realtime = {
-//   type: "realtime";
-//   group?: string;
-//   updateInterval: number; // 1000 * 60 -> 1m
-// };
+type Realtime = FieldBase<"realtime"> & {
+  updateInterval?: number; // 1000 * 60 -> 1m
```
KaWaite, line 69: You set this and the StyleCode's src in the component declaration. If these will always have values/need values please type them as necessary.

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/ButtonLink.tsx
```diff
@@ -20,10 +20,15 @@ const ButtonLink: React.FC<BaseFieldProps<"buttonLink">> = ({ value, editMode, o
 
   const handleChangeButtonLink = useCallback(
     (e: React.ChangeEvent<HTMLInputElement>) => {
-      setLink(e.currentTarget.value);
+      let url = e.currentTarget.value;
+      const prefix = "http://";
+      if (!url.match(/^[a-zA-Z]+:\/\//)) {
+        url = prefix + url;
```
KaWaite, line 26: HAve you tested this? Also, most websites are https now, right? Perhaps https:// is better to attach. Though I am a little unsure this is good/needed. I feel like its prone to unexpected behavour

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
```diff
@@ -0,0 +1,150 @@
...
+    [enableUpdate, getTimeRemaining, intervalInSecond],
+  );
+  const clearTimer = useCallback(() => {
+    if (Ref.current) clearInterval(Ref.current);
+  }, []);
+
+  const handleChangeupdateState = useCallback(() => {
+    changeUpdateState(!enableUpdate);
+    setTimer("00:00:30");
+    if (enableUpdate) startTimer(intervalInSecond);
```
KaWaite, line 50: Since you are doing !enableUpdate above, I think this will start the timer when enableUpdate is false because if it is **becoming** true, in this if statement it will still be false until the next time it is called.

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
```diff
@@ -0,0 +1,150 @@
...
+        updateInterval: parseInt(e.currentTarget.value),
+      });
+      startTimer(intervalInSecond);
+    },
+    [intervalInSecond, onUpdate, startTimer, value],
+  );
+
+  useEffect(() => {
+    if (enableUpdate) {
+      setTimer("00:00:30");
```
KaWaite, line 67: This is assuming the default 30s. But if its editted to be a diffferent value this will be wrong

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
```diff
@@ -0,0 +1,150 @@
...
+      startTimer(intervalInSecond);
```
KaWaite, line 68: You are doing almost the same thing in this useEffect AND in the usecallback that is handling the state change. I think it makes sense to simplify the handleChangeupdateState function to only handle that state change and allow this useeffect to handle the rest (as it is doing now)

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
```diff
@@ -0,0 +1,150 @@
...
+      const id = setInterval(() => {
+        if (enableUpdate) {
+          getTimeRemaining(interval);
+          interval--;
+          if (interval === 0) interval = intervalInSecond;
+        }
+      }, 1000);
+
+      Ref.current = id;
+    },
```
KaWaite, line 40: To be honest I can't tell if this'll work or not. Is this tested? If its tested and working, the only thing that would be nice is the const id be renamed to perhaps remainingTime, or just added directly like Ref.current = setInterval.....

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
```diff
@@ -0,0 +1,150 @@
...
+            value={intervalInSecond}
+            onChange={handleChangeUpdateTime}
+          />
+        </FieldValue>
+      </Field>
+    </Wrapper>
+  ) : (
+    <Wrapper>
+      <Field gap={10}>
+        <Icon icon="clock" size={24} />
```
KaWaite, line 88: This is a static icon. What we want here is a dynamic icon that updates with the time. Does Antd have this in their icon set or loading section?

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/ButtonLink.tsx
nina992: if I didn't add ths validation when I set the link to "www.google.com " ,  causes the browser to assume a relative path and append localhost to the url , I tested that and it works fine

nina992: I added validation for the default value also

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/types.ts
nina992: Done

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: but starting the timer is set only when enableUpdate is true

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: if this edited , I will check if there is any previous timer is working then I stop it and start new timer with the new edited value

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: yes it is tested , and done for the renaming

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: Antd icons  have spinning property but for the custom icons don't I can try to add rotation property to icon , or we can change the icon and use existed icons in antd , what do you think ?

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/types.ts
nina992: c9e5706fc4bb3535ff518c320d648fb4f2ea4b1b

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/ButtonLink.tsx
nina992: c9e5706fc4bb3535ff518c320d648fb4f2ea4b1b

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: c9e5706fc4bb3535ff518c320d648fb4f2ea4b1b

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: c9e5706fc4bb3535ff518c320d648fb4f2ea4b1b

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: done c9e5706fc4bb3535ff518c320d648fb4f2ea4b1b

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: c9e5706fc4bb3535ff518c320d648fb4f2ea4b1b

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/ButtonLink.tsx
KaWaite: I think it is best to just do the regex check on the link in the onClick with the window.open.
We shouldn't be changing the saved value, because the user would get confused, and we only really need the https:// attached before opening the page.

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
KaWaite: That is exactly what I mean. In this handleChangeUpdateState function, the first thing you do is `changeUpdateState(!enableUpdate);`. This, if enableUpdate was false, will make it true, right? But it won't be true in this useCallback until the next time the useCallback is used. Which means the enableUpdate in ->`if(enableUpdate) startTimer(intervalInSecond)` is still false. 
Understand?

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/ButtonLink.tsx
nina992: done & tested

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: you mean I should change the state `changeUpdateState(!enableUpdate); `after checking the `if(enableUpdate)startTimer(intervalInSecond)` but even if I do that there will be a ms while checking the current state , there is another solution to use Ref for example , is that what you mean ?

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Realtime.tsx
nina992: @lavalse  hello , we need your support here , in realtime update I used static icon that you provided us with , but shouldn't be with animation ? if yes , would you please provide us the library you're using

KaWaite: @nina992 Please rely on Slack for communication between core members. @lavalse MIGHT notice this message on github, but Slack is a much more reliable place for quick replies.

nina992: sure I did mention red-san on slack

KaWaite (approved):

PR #118, opened 2023-02-16
fix(plugin): field's move up down functionality

nourbalaha:
Done:
- Implement up and down functionality for `PointColor` and `PointStroke`
- Keep state for common components
- Add `swap`, `moveItemUp`, `moveItemDown` and `removeItem` methods into `util.js`

### DIFF
--- a/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/point/common/ColorField.tsx
+++ b/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/point/common/ColorField.tsx
@@ -1,6 +1,6 @@
 import { Icon } from "@web/sharedComponents";
 import { styled } from "@web/theme";
-import { ChangeEvent, useState } from "react";
+import { ChangeEvent, useEffect, useState } from "react";
 
 import { FieldTitle, FieldValue, FieldWrapper, TextInput } from "../commonComponents";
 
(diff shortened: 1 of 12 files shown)


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/point/common/ColorField.tsx
```diff
@@ -19,6 +19,11 @@ const ColorField: React.FC<Props> = ({ title, titleWidth, color, onChange }) =>
   const [text, setText] = useState(color);
   const [selectedColor, setSelectedColor] = useState(color);
 
+  useEffect(() => {
+    setText(color);
+    setSelectedColor(color);
```
KaWaite, line 24: Is there any reason you need both text and selectedColor? It looks like this component uses both always together, right?

KaWaite (approved):

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/point/common/ColorField.tsx
nourbalaha: Text is for whatever is written in the text box while selected color is only set when a valid color is entered.
Screenshot 0005-02-20 at 10 32 24 AM

PR #121, opened 2023-02-17
feat(plugin): add building filter

keiya01:


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/index.ts
```diff
@@ -40,6 +41,7 @@ const fields: Fields<FieldComponent> = {
   // polygon
   // 3d-model
   clipping: { Component: Clipping, hasUI: false },
+  buildingFilter: { Component: BuildingFilter, hasUI: false },
```
KaWaite, line 44: This should be hasUI: true. 
Please check the design again. The ui on the right is the settings UI (in this case, it is empty), on the left is the user facing UI.
The settings UI for this component and a few others removes the arrow icon and border bottom. 

I'll do that in a different PR, so please just make it so the UI you have made in this PR is shown not inEditor, but for the user

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingFilter/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+        {
+          border: "2px solid #00BEBE",
+        },
+        {
+          border: "2px solid #00BEBE",
+        },
+      ] as CSSProperties[],
+    }),
+    [],
+  );
```
KaWaite, line 41: Can't this just be moved outside the component instead of in a useMemo?

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingFilter/hooks.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (value && Array.isArray(value)) {
+        handleUpdate(prop, value as [from: number, to: number]);
+      }
+    },
+    [handleUpdate],
+  );
+
+  useEffect(() => {
+    setOptions({ ...value });
+  }, [value, onUpdate]);
```
KaWaite, line 42: I don't think you want `options` to be set to value, right?
  `options` type is value's type minus a few fields right? 
Also, won't this useEffect make any updates call setOptions twice (once in the handleUpdate function and then again here in the useEffect when the value change triggers it)??

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/index.ts
keiya01: I fixed.
[fix: hasUI to true](https://github.com/eukarya-inc/reearth-plateauview/pull/121/commits/b29d629359b648eade13622c6b926fbeb5ae18ce)

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingFilter/index.tsx
keiya01: I fixed.
[fix: cull out styleProps to out of component](https://github.com/eukarya-inc/reearth-plateauview/pull/121/commits/5746fd20476a11ce24396a9674ca2c61198e51a3)

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingFilter/hooks.ts
keiya01: I fixed.
[fix: prevent unnecessary render](https://github.com/eukarya-inc/reearth-plateauview/pull/121/commits/44fd8f17ea04a3832dedcc71dae046af0b73de61)

KaWaite (approved):

PR #125, opened 2023-02-17
feat:(plugin) add building color

keiya01:


plugin/src/sidebar.ts
```diff
@@ -394,7 +410,16 @@ reearth.on("message", ({ action, payload }: PostMessageProps) => {
     });
   }
 
-  // For 3dtiles color
+  // FIXME(@keiya): We need to compose transparency with color,
+  //                but currently rgba is not working on NLS.
+  //                So I will fix NLS and implement composing color.
```
keiya01, line 415: I will fix this in other PR because this need to fix NLS.
Currently, composing transparent color with 3dtiles color is not working. For example, when I change transparency to 0.5, also change building color to blue, then the building color will be opaque.

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingColor/conditions.ts
```diff
@@ -0,0 +1,93 @@
...
+const FIREPROOF_CONDITIONS: Condition[] = [
+  [equalString(conditionalFireproof, "耐火"), "color('#7c7b87')"],
+  [equalString(conditionalFireproof, "準耐火造"), "color('#8e9bb0')"],
+  [equalString(conditionalFireproof, "その他"), "color('#d083af')"],
+  [equalString(conditionalFireproof, "不明"), "color('#89c2ea')"],
+  DEFAULT_CONDITION,
+];
+
+export const makeSelectedFloodCondition = (
+  propertyName: string | undefined,
```
KaWaite, line 73: Are these hardcoded values what we want? (Did @rot1024 say to do it like this?)

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingColor/conditions.ts
```diff
@@ -0,0 +1,93 @@
...
+];
+
+const conditionalFireproof = "建物利用現況_耐火構造種別";
+const FIREPROOF_CONDITIONS: Condition[] = [
+  [equalString(conditionalFireproof, "耐火"), "color('#7c7b87')"],
+  [equalString(conditionalFireproof, "準耐火造"), "color('#8e9bb0')"],
+  [equalString(conditionalFireproof, "その他"), "color('#d083af')"],
+  [equalString(conditionalFireproof, "不明"), "color('#89c2ea')"],
+  DEFAULT_CONDITION,
+];
```
KaWaite, line 70: Are these hardcoded values what we want? (Did @rot1024 say to do it like this?)

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingColor/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+      </StyledRadio>
+      <StyledRadio value="purpose">
+        <Label>用途による塗分け</Label>
+      </StyledRadio>
+      <StyledRadio value="structure">
+        <Label>建物構造による塗分け</Label>
+      </StyledRadio>
+      <StyledRadio value="fireproof">
+        <Label>耐火構造種別による塗分け</Label>
+      </StyledRadio>
```
KaWaite, line 36: I don't think all 3d tiles that PLATEAU has will have every single type. I could be wrong though. Is this just temporarily hardcoded?

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingColor/conditions.ts
keiya01: I discussed with @lavalse. I decided we need to do this for now,  but we should have more efficient way.

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingColor/index.tsx
keiya01: Oops... I didn't notice it. I will fix it.

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingColor/conditions.ts
lavalse: Yes. I agree we need a more efficient way. But I think it is ok now to do this.

KaWaite (approved):

PR #148, opened 2023-02-21
feat(plugin): field's multiple conditions

nourbalaha:
Done:
- Add functionality to point color condition component
- Add functionality to point stroke condition component
- Add functionality to polygon color condition component
- Add functionality to polygon stroke condition component
- Add functionality to polyline color condition component


KaWaite (approved):
Approved but will do more testing before merge

PR #150, opened 2023-02-22
feat(plugin): support csv field

keiya01:
Screen Shot 2023-02-22 at 10 15 00


plugin/src/sidebar.ts
```diff
@@ -367,6 +368,39 @@ reearth.on("message", ({ action, payload }: PostMessageProps) => {
...
+    const { dataID } = payload;
+    const layerId = addedDatasets.find(ad => ad[0] === dataID)?.[2];
+    const layer = reearth.layers.findById(layerId);
+    reearth.ui.postMessage({
+      action,
+      locationNames: getLocationNamesFromFeatureProperties({
+        ...(layer.computed?.features[0] || {}),
+      }),
+    });
+  }
```
KaWaite, line 402: Can you comment this out too since it isn't being used yet (the useEffect with the ToDo comment seems to be the only thing using this)

KaWaite (approved):

PR #152, opened 2023-02-22
fix(plugin): set scene shadow

keiya01:


plugin/src/sidebar.ts
```diff
@@ -44,6 +44,7 @@ const defaultProject: Project = {
         tile_type: "url",
       },
     ],
+    atmosphere: { shadows: true },
```
KaWaite, line 47: There is another defaultProjects that you'll need to update too. Please search for it and update there too (I think its in the sidebar's main hooks file)

keiya01: [fix](https://github.com/eukarya-inc/reearth-plateauview/pull/152/commits/439ddec873690071a5f7fb0cc59d7d0abe7044ec)

KaWaite (approved):

PR #171, opened 2023-02-24
fix(plugin): close building search when dataset delete

airslice:
Building search popup should be closed when its dataset been deleted.


plugin/src/sidebar.ts
```diff
@@ -224,9 +224,15 @@ reearth.on("message", ({ action, payload }: PostMessageProps) => {
     reearth.layers.delete(addedDatasets.find(ad => ad[0] === payload)?.[2]);
     const idx = addedDatasets.findIndex(ad => ad[0] === payload);
     addedDatasets.splice(idx, 1);
+    if (openedBuildingSearchDataID && openedBuildingSearchDataID === payload) {
+      reearth.popup.close();
+    }
```
KaWaite, line 229: Should you not set openedBuildingSearchDataID to null here?

plugin/src/sidebar.ts
```diff
@@ -224,9 +224,15 @@ reearth.on("message", ({ action, payload }: PostMessageProps) => {
...
   } else if (action === "removeAllDatasetsFromScene") {
     reearth.layers.delete(...addedDatasets.map(ad => ad[2]));
     addedDatasets = [];
+    if (openedBuildingSearchDataID) {
+      reearth.popup.close();
+    }
```
KaWaite, line 235: Should you not set openedBuildingSearchDataID to null here?

KaWaite (approved):

PR #172, opened 2023-02-24
fix(plugin): name for field components

nourbalaha:
Done:
- Add more info to field components' name
- Display full name with long names


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/index.tsx
```diff
@@ -174,6 +207,8 @@ const BodyWrapper = styled(AccordionItemPanel)`
 const Title = styled.p`
   margin: 0;
   user-select: none;
+  width: 200px;
```
KaWaite, line 210: Is this 200px going to work in mobile as well?

nourbalaha: I think 200px is good for mobile as well. setting the width is necessary to make the paragraph take multiple lines

KaWaite: What if they have a smaller phone like an iphone 7 or iphone se? Is 200px going to be wider than their screen?

nourbalaha: iPhone 7's width is 375px and iPhone SE's width is 320px right?

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/index.tsx
```diff
@@ -96,7 +122,11 @@ const FieldComponent: React.FC<Props> = ({
                     {Field && (
                       <ArrowIcon icon="arrowDown" size={16} direction="right" expanded={expanded} />
                     )}
-                    <Title>{field.type === "template" ? field.name : fieldName[field.type]}</Title>
+                    <Title>
+                      {field.type === "template"
+                        ? `${field.name}(${getFieldGroup(field.type)})`
+                        : `${fieldName[field.type]}(${getFieldGroup(field.type)})`}
```
KaWaite, line 128: Can you build the name in a useMemo so it doesn't have to be calculated each render

KaWaite (approved):

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/index.tsx
nourbalaha: Done!

PR #175, opened 2023-02-24
fix(plugin): description markdown's image size

nourbalaha:
Done:
- Fix description markdown's image size

### DIFF
--- a/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Description.tsx
+++ b/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Description.tsx
@@ -73,6 +73,6 @@ const SwitchWrapper = styled.div`
 
 const Markdown = styled.div`
   img {
-    width: 100%;
+    max-width: 100%;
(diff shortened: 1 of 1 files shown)


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/Description.tsx
```diff
@@ -70,3 +70,9 @@ const SwitchWrapper = styled.div`
   gap: 10px;
   margin-top: 12px;
 `;
+
+const Markdown = styled.div`
+  img {
+    width: 100%;
```
KaWaite, line 76: Is it possible to do max-width: 100% and still successfully fix this issue? 
The problem with setting width to 100% is that ALL images, all the time will be 100%, even if the image is smaller than the width of the DatasetCard

nourbalaha: Makes sense! Done!

KaWaite (approved):

PR #180, opened 2023-02-28
perf(plugin): improve about data button's performance

nourbalaha:
Done:
- Adding keys to data catalog's file tree nodes
- Keeping track of expanded keys
- Using expanded keys to implement about data button

Not done:
- For some strange reason about data button does not work on data that has already been published before even if it becomes unpublished again


plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/Folder.tsx
```diff
@@ -19,30 +18,20 @@ const Folder: React.FC<Props> = ({
...
-      if (!("children" in item)) return "dataID" in item && item.dataID === selectedID;
-      return item.children.some(child => isChildSelected(child));
-    },
-    [selectedID],
-  );
-
   useEffect(() => {
-    open(() => expandAll || (!!selectedID && isChildSelected(item)));
-  }, [expandAll, isChildSelected, item, selectedID]);
+    open(() => expandAll || expandedKeys.includes(nodeKey));
```
KaWaite, line 29: This will happen for ALL folders EVERYTIME a key is added to expandedKeys, right?
Please add an if statement `if(expanded || expandedKeys.includes(nodeKey)` so that we don't update the isOpen state so much.

nourbalaha: I am not sure whether this is necessary or not.
Do you mean something like this:
```ts
  useEffect(() => {
    expandAll || expandedKeys.includes(nodeKey)
      ? open(expandAll || expandedKeys.includes(nodeKey))
      : open(false);
  }, [expandAll, expandedKeys, nodeKey]);
```

nourbalaha: Done!

KaWaite (approved):

PR #185, opened 2023-03-01
fix(plugin): override strategy for 3dtiles

keiya01:


plugin/web/extensions/sidebar/core/components/hooks.ts
```diff
@@ -690,9 +690,13 @@ export const processOverrides = (components?: FieldComponent[], cleanseOverride?
     }
     return;
   }
+
+  const sortedComponents = [...components].sort(
+    (a, b) => (a.updatedAt?.getTime() ?? 0) - (b.updatedAt?.getTime() ?? 0),
+  );
```
keiya01, line 696: I added updatedAt property to override property with correct order.
For example, 3dtiles has transparency and building color field, and these field override `color` property on 3dtiles layer. So color property that is overridden by transparency field is always overridden by color building field, because the order of transparency field is before building color field in `DatasetComponents`.

KaWaite: Doesn't the order depend on what order the user adds the field components?
Also, @lavalse and I have discussed a couple times and decided that the last component added should take priority. For example, if you have FieldComponentA that changes point color to Red, and then add FieldComponentB after that which changes point color to Blue, the point should be Blue.

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/3dtiles/BuildingTransparency/useBuildingTransparency.ts
```diff
@@ -1,94 +1,85 @@
...
+            }
+            const composedRGBA = [...rgba.slice(0, -1), transparency] as RGBA;
+            return [k, rgbaToString(composedRGBA)];
+          }),
+        },
+      };
+    })();
+
+    onUpdateRef.current?.({
+      color: expression,
```
KaWaite, line 80: Should this not be something like `{"3dtiles": {color: expression}}` or something?

KaWaite (approved):

PR #208, opened 2023-03-07
feat(plugin): show description for czml datasets

nourbalaha:
- Show `description` field content in HTML block extension for czml datasets.


plugin/src/sidebar.ts
```diff
@@ -555,6 +555,25 @@ function createLayer(dataset: DataCatalogItem, overrides?: any) {
...
+                property: {
+                  default: {
+                    html: dataset.desc,
+                  },
+                },
+              },
+            ],
+            property: {
+              default: {
+                showTitle: false,
```
KaWaite, line 573: Please make this true

KaWaite (approved):

plugin/src/sidebar.ts
nourbalaha: Done!

PR #210, opened 2023-03-07
fix(plugin): usecase datasets' name

nourbalaha:
Done:
- Change the name of usecase datasets use to be "zzz" instead of "xxx/yyy/zzz"

### DIFF
--- a/plugin/web/extensions/sidebar/core/components/hooks.ts
+++ b/plugin/web/extensions/sidebar/core/components/hooks.ts
@@ -90,20 +90,11 @@ export default () => {
     postMsg({ action: "init" }); // Needed to trigger sending initialization data to sidebar
   }, []);
 
-  // getNameFromPath("xxx/yyy/zzz") -> zzz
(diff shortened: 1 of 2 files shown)


plugin/web/extensions/sidebar/core/components/hooks.ts
```diff
@@ -90,11 +90,20 @@ export default () => {
...
+  };
+
   useEffect(() => {
     const catalogBaseUrl = catalogURL || backendURL;
     if (catalogBaseUrl) {
       getDataCatalog(catalogBaseUrl, catalogProjectName).then(res => {
-        setCatalog(res);
+        const catalog = res.map(item => {
+          return { ...item, name: getNameFromPath(item.name) };
+        });
```
KaWaite, line 105: The catalog is very large, so please either add the getNameFromPath inside the getDataCatalog. I know that code is a bit complex, so if you have issues let me know and I'll try to help

nourbalaha: Done!

KaWaite (approved):

PR #221, opened 2023-03-07
fix(plugin): style code component

basel-i:
## Overview

This PR uses component style code override.

## Test

Test project [here](https://test.reearth.dev/edit/01gty4xw1ysb8smcdycxp9qtm7).
Test shared project [here](https://test.reearth.dev/published.html?alias=ejdgeebgic). (建築物モデル（中央区)

## Screen Shot
Screenshot 2023-03-07 at 4 55 32 PM

Screenshot 2023-03-07 at 5 19 26 PM

### DIFF
--- a/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/StyleCode/hooks.ts
+++ b/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/StyleCode/hooks.ts
@@ -7,19 +7,19 @@ export default ({ value, onUpdate }: Pick<BaseFieldProps<"styleCode">, "value" |
   const [code, editCode] = useState(value.src);
 
   const onApply = useCallback(() => {
-    try {
-      const styleObject = JSON5.parse(code);
-      onUpdate({
-        ...value,
-        src: code,
-        override: styleObject,
-      });
-      // eslint-disable-next-line no-empty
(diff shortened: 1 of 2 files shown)


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/StyleCode/hooks.ts
```diff
@@ -1,48 +1,32 @@
...
         ...value,
-        src: e.target.value,
+        src: code,
+        override: styleObject,
       });
-    },
-    [onUpdate, value],
-  );
+      // eslint-disable-next-line no-empty
+    } catch (error) {}
```
KaWaite, line 18: Since you are not catching any error I dont think they try/catch is needed.

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/StyleCode/hooks.ts
```diff
@@ -1,48 +1,32 @@
...
+  }, [onUpdate, code, value]);
 
   useEffect(() => {
-    updateStyle(value.src);
+    onApply();
```
KaWaite, line 22: Is this for initialization? If so, it shouldn't be needed since you pass an override field in the onUpdate.

PR #228, opened 2023-03-08
feat(plugin): enable Clamp to ground property

nina992:
# Overview 
We need to enable Clamp to ground property for polygon, marker and models.


plugin/src/sidebar.ts
```diff
@@ -584,12 +584,14 @@ function createLayer(dataset: DataCatalogItem, overrides?: any) {
             style: "point",
             pointSize: 10,
             pointColor: "white",
+            clampToGround: true,
```
keiya01, line 587: About marker, we need to use heightReference.

```suggestion
            heightReference: "clamp",
```

keiya01 (commented):
Also I think we need to set clampToGround setting to CZML in [this line](https://github.com/eukarya-inc/reearth-plateauview/blob/3fcfc802514590ba4d8c6f2300e79578cff64b58/plugin/src/sidebar.ts#L607) like the following.

```js
{
  resource: {},
  marker: { heightReference: "clamp" },
  polyline: { clampToGround: true },
  polygon: { clampToGround: true },
}
```

plugin/src/sidebar.ts
nina992: Done

keiya01 (approved):

plugin/src/sidebar.ts
nina992: 83e66e860c7ad90109b6c1185f6cdcbb085c5d10

PR #229, opened 2023-03-08
feat: infobox with new design

airslice:
## Overview

This PR recreated the infobox with new design.

## Features

UI:
- When `inEditor` the infobox has both view panel and edit panel.
- When in published page the infobox has only view panel.

Properties:
- Now properties of 3d tiles feature are split into 3 parts:
  - Part I: 15 common properties (which can be edit in edit panel)
  - Part II: floor properties (actually = all properties - common properties - override apperence properties - attributes)
  - Part III: `attributes` (currently using react-json-view to render the tree)
- Only show exist properties on view panel.

## Test

Test project [here](https://test.reearth.dev/edit/01gsvnnzjawsjf6pjzt4cj262k).

## Screenshot
test reearth dev_edit_01gsvnnzjawsjf6pjzt4cj262k(1920 desktop)

test reearth dev_edit_01gsvnnzjawsjf6pjzt4cj262k(1920 desktop) (1)

test reearth dev_edit_01gsvnnzjawsjf6pjzt4cj262k(1920 desktop) (2)


plugin/web/extensions/infobox/core/components/editPanel/FieldsEditor.tsx
```diff
@@ -0,0 +1,109 @@
...
+  fields,
+  isSaving,
+  onFieldCheckChange,
+  onFieldTitleChange,
+  onFieldMove,
+  saveTemplate,
+}) => {
+  const onSave = useCallback(() => {
+    saveTemplate();
+  }, [saveTemplate]);
```
KaWaite, line 30: Are these kind of callbacks really necessary? 
I think this just makes React do more work for no real benefit since there is no additional calculations other than the one function.
If I am wrong, let me know. But I think callbacks like this will just add overhead

plugin/web/extensions/infobox/core/components/editPanel/FieldsEditor.tsx
```diff
@@ -0,0 +1,109 @@
+// import TEST_FIELDS from "@web/extensions/infobox/TEST_FIELDS";
```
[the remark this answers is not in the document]
airslice: Sorry i missed this one...

plugin/web/extensions/infobox/core/components/editPanel/FieldsEditor.tsx
airslice: Yes you're right. There was some other logic here, after some refactor the save is implemented in parent hooks so it should be removed here.

KaWaite (approved):

PR #245, opened 2023-03-10
fix(plugin): current time field

keiya01:


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/general/CurrentTime.tsx
```diff
@@ -39,7 +39,20 @@ const CurrentTime: React.FC<BaseFieldProps<"currentTime">> = ({
...
+            .split(/-|\//)
+            .map(s => s.padStart(2, "0"))
+            .join("-")
+        : undefined;
+      const formattedTime = options.time
+        ? options.time
+            .split(/:/)
+            .map(s => s.padStart(2, "0"))
+            .join(":")
+        : undefined;
```
keiya01, line 53: This code format date string like `2011-3-11` or `12:0:0` to `2011-03-11` or `12:00:00`. Because JS `Date` API can't parse `2011-3-11T12:0:0`.

KaWaite (approved):

PR #247, opened 2023-03-10
fix(plugin): building search override

airslice:
## Overview

Sidebar has an override system for field components, but building search was override the 3d tiles' style independently.

This PR connect the building search's override into the override system.

## Test 

Test project [here](https://test.reearth.dev/edit/01gv5fejanm86awez4zmmnvqn1).

## Scrren shot 

https://user-images.githubusercontent.com/21994748/225013390-d73597d2-eebb-4824-b1d6-6b321500ab75.mp4


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/hooks.ts
```diff
@@ -40,10 +47,11 @@ export default ({
   }, [selectedGroup, dataset.components]); // eslint-disable-line react-hooks/exhaustive-deps
 
   useEffect(() => {
-    if (activeComponentIDs) {
-      onOverride?.(dataset.dataID, activeComponentIDs);
+    const buildingSearchActive = buildingSearch?.find(b => b.dataID === dataset.dataID)?.active;
+    if (activeComponentIDs || buildingSearchActive) {
+      onOverride?.(dataset.dataID, [...(activeComponentIDs ? activeComponentIDs : [])]);
```
KaWaite, line 52: I can't decide if it is good or not, but these days I've been shortening this kind of situation to something like
`...(activeComponentIDs ?? [])`. 
You don't have to change here, but just in case you like shorter code :P

airslice: Seems better! I updated this.

PR #255, opened 2023-03-13
feat(plugin): save data catalog's open folders

nourbalaha:


plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
```diff
@@ -30,6 +30,17 @@ const FileTree: React.FC<Props> = ({
...
 
+  useEffect(() => {
+    const { expandedKeys } = window as any;
+    if (expandedKeys) {
+      const newExpandedKeys = [...expandedKeys];
+      setExpandedKeys(newExpandedKeys);
+    }
+    setTimeout(() => {
+      (window as any).expandedKeys = undefined;
+    }, 500);
```
KaWaite, line 41: A couple questions:
- Why does this useEffect's deps array have two functions not used in the useEffect?
- We discussed on slack, but you shouldn't be able to access the window object from a plugin, so this expandedKeys should not be accessible (if there is anything to discuss around this, please do on slack)

plugin/web/extensions/sidebar/modals/datacatalog/index.html
```diff
@@ -23,7 +23,11 @@
 window.addEventListener("message", function (e) {
     if (e.source !== parent) return;
     if (e.data.action === "catalogModalOpen" && e.data.payload) {
-        if(e.data.payload.dataset) window.selectedDataset = e.data.payload.dataset;
+        if(e.data.payload.dataset) {
+            window.selectedDataset = e.data.payload.dataset;
+        } else if(e.data.payload.expandedKeys) {
+            window.expandedKeys = e.data.payload.expandedKeys;
```
KaWaite, line 29: Again, window shouldn't be accessible from the plugin.

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
nourbalaha: Removed the two functions that are not used in the useEffect

nourbalaha: Done!

plugin/web/extensions/sidebar/modals/datacatalog/index.html
nourbalaha: Done!

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
```diff
@@ -23,13 +24,49 @@ const FileTree: React.FC<Props> = ({
...
-  const handleSelect = useCallback((dataID?: string) => {
-    select(dataID);
+  const handleSelect = useCallback((item?: DataCatalogItem) => {
+    selectItem(item);
   }, []);
 
+  useEffect(() => {
+    postMsg({ action: "catalogModalOpen" }); // Needed to trigger sending client storage data from Sidebar
+  }, []);
+
```
KaWaite, line 37: I don't think this is really good.
Already this filetree has become really heavy (we will talk about that later). But this catalogModalOpen is going to send the datacatalog the catalog on open twice. Not sure why it is needed here.

KaWaite: I think you might want to either make a new action, or update the current catalogModalOpen action to be able to avoid sending the catalog each time.
Using this action for dataset and the folder keys separately is fine because they can be empty and/or small. But I feel like the catalog should be avoided except for when necessary

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
```diff
@@ -23,13 +24,49 @@ const FileTree: React.FC<Props> = ({
...
+            );
+            if (index >= 0) expandedFolders.splice(index, 1);
+            setExpandedFolders(expandedFolders);
+          }
+          setTimeout(() => {
+            postMsg({
+              action: "saveDataset",
+              payload: { dataset: undefined },
+            });
+          }, 500);
```
KaWaite, line 60: I'm not sure we should remove the selected dataset with a timeout. Perhaps on modal close (in the sidebar.ts file) is better

nourbalaha: on modal close is not triggered when you click outside the modal, it is only triggered when you click on the close button.
However I removed the setTimeout and just left the postMsg method.
What do you think?

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
nourbalaha: You've already did something like this here:
https://github.com/eukarya-inc/reearth-plateauview/blob/a8c27dcbfaed3bbaf7d6213e154d714cb4ff5a9c/plugin/web/extensions/sidebar/modals/datacatalog/components/hooks.ts#L36
I feel like you are mixing between `initDataCatalog` and `catalogModalOpen`

KaWaite: Nour that is exactly what I am tlaking about. Because we do that initDataCatalog, which receives catalog, among other things, I don't see why we need this catalogModalOpen in the Filetree.tsx file. Can you explain why this is desirable here?

nourbalaha: Because, otherwise the expandedFolders and dataset variables will not be received even if I get them from a new action in the sidebar.

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/FileTree.tsx
KaWaite: I think our api has onClose where you can pass a callback

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/Folder.tsx
```diff
@@ -26,13 +27,13 @@ const Folder: React.FC<Props> = ({
   const [isOpen, open] = useState(false);
 
   useEffect(() => {
-    if (expandAll || expandedFolders.find(item => (item.id ? item.id === id : item.name === name)))
+    if (expandAll || expandedFolders?.find(item => (item.id ? item.id === id : item.name === name)))
```
KaWaite, line 30: Is there any chance the item doesn't have an ID? Are you not adding an ID to all catalog groups/items when processing the catalog (I think in the api.ts file??)??
If I am correct, do we really need item.name? We might be able to simplify expandedFolders to be just an array of IDs?

nourbalaha: In order to add functionality to the about data button, I use the `item.path` to expand the parent folders of the selected dataset. 
`item.path` is an array of folder names which is `item.name` and it doesn't have the item's id. that is why I expand using the `item.id` and if there is no id then I use the `item.name`.
I hope that makes sense.

plugin/web/extensions/sidebar/modals/datacatalog/components/hooks.ts
```diff
@@ -43,6 +54,23 @@ export default () => {
         setAddedDatasetDataIDs(e.data.payload.addedDatasets);
         setCatalog(e.data.payload.catalog);
         setEditorState(e.data.payload.inEditor);
+        setExpandedFolders(e.data.payload.expandedFolders);
```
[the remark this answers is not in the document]
nourbalaha: Done!

plugin/src/sidebar.ts
```diff
@@ -255,6 +257,10 @@ reearth.on("message", ({ action, payload }: PostMessageProps) => {
     }
   } else if (action === "triggerCatalogOpen") {
     reearth.ui.postMessage({ action });
+  } else if (action === "saveExpandedFolders") {
+    expandedFolders = [...payload.expandedFolders];
+  } else if (action === "saveDataset") {
+    dataset = { ...payload.dataset };
```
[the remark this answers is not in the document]
nourbalaha:
```
let expandedFolders: { id?: string; name?: string }[] = [];
let dataset: DataCatalogItem | undefined = undefined;
```

About expandedFolders:
expandedFolders is initialized with an empty array so it will not be undefined.

About dataset
Dataset's type is `DataCatalogItem | undefined`
Please remember that dataset can be set to undefined for example when resetting the the dataset after clicking the about button. so there is not need to check if it is undefined or not.

pyshx (approved):

PR #263, opened 2023-03-15
feat(plugin): datasets reordering by drag and drop

nourbalaha:
Done:
- Add support for reordering datasets by adding drag and drop functionality

### DIFF
--- a/plugin/web/extensions/sidebar/core/components/hooks/projectHooks.ts
+++ b/plugin/web/extensions/sidebar/core/components/hooks/projectHooks.ts
@@ -193,6 +193,17 @@ export default ({
     postMsg({ action: "removeAllDatasetsFromScene" });
   }, []);
 
+  const handleProjectDatasetsUpdate = useCallback((datasets: DataCatalogItem[]) => {
+    updateProject(({ sceneOverrides }) => {
+      const updatedProject = {
+        sceneOverrides,
+        datasets,
+      };
+      postMsg({ action: "updateProject", payload: updatedProject });
+      return updatedProject;
+    });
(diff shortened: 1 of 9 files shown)


plugin/web/extensions/sidebar/popups/mobileDropdown/index.tsx
KaWaite, line 1: We don't really support editMode on mobile, so this isn't really necessary

plugin/web/extensions/sidebar/core/components/hooks/projectHooks.ts
```diff
@@ -193,6 +193,17 @@ export default ({
...
+  const handleProjectDatasetsUpdate = useCallback((datasets: DataCatalogItem[]) => {
+    updateProject(({ sceneOverrides }) => {
+      const updatedProject = {
+        sceneOverrides,
+        datasets,
+      };
+      postMsg({ action: "updateProject", payload: updatedProject });
+      return updatedProject;
+    });
+  }, []);
```
KaWaite, line 205: This isn't necessary since we have the handleDatasetUpdate function in the datasetHooks.
BTW, this handleProjectDatasetUpdate is problematic because it only updates the project that is used for UI. It is missing updating overrides and other things that are necessary when a dataset is updated

plugin/web/extensions/sidebar/popups/mobileDropdown/index.tsx
nourbalaha: Done!

KaWaite (approved):

PR #273, opened 2023-03-17
feat(plugin): color gradient component

nourbalaha:


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/point/PointColorGradient.tsx
```diff
@@ -1,13 +1,163 @@
...
+    [colorGradient],
+  );
+
+  const conditions = useMemo(() => {
+    const { field, min, max, step, startColor, endColor } = colorGradient;
+    return generateConditions(field, min, max, step, startColor, endColor);
+  }, [colorGradient]);
+
+  useEffect(() => {
+    if (!isActive || isEqual(omit(colorGradient, "override"), omit(value, "override"))) return;
```
KaWaite, line 100: Can you update the colorGradient useState to be just the color, not value. Value includes override and type and id, etc.
If you do that, you can more simply and deliberately check if the two are equal

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/point/PointColorGradient.tsx
```diff
@@ -1,13 +1,163 @@
...
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      const max = !isNaN(parseFloat(e.currentTarget.value)) ? parseFloat(e.currentTarget.value) : 1;
+      setColorGradient({ ...colorGradient, max });
+    },
+    [colorGradient],
+  );
+
+  const handleStartColorUpdate = useCallback(
+    (color: string) => {
+      // colors should be hex for now
```
KaWaite, line 66: Can you look in to [this page](https://www.w3schools.com/jsref/jsref_getcomputedstyle.asp) for perhaps supporting red, green, pink, etc?

nourbalaha: I saw this yesterday but I thought that we can not use window object.
```ts
window.getComputedStyle(element, null);
```
I will check again.

KaWaite (approved):

PR #277, opened 2023-03-17
fix(plugin): clampToGround

pyshx:


plugin/src/sidebar.ts
```diff
@@ -658,3 +668,16 @@ function createLayer(dataset: DataCatalogItem, overrides?: any) {
       : { ...(overrides ?? {}) }),
   };
 }
+
+const defaultOverrides = {
```
KaWaite, line 672: Can you use this everywhere that it is appropriate? From what I can see easily, when format is czml we want to use all these values as default. If there is anywhere else too please use these values to keep things consistent

PR #280, opened 2023-03-17
fix(plugin): multiple folders open at the same times on data catalog

ghost:
# Overview
Multiple folders open at the same time on the data catalog when they have the same name

# Problem
The issue is that not all catalog items have unique value or an `id` property

# Solution
Create a unique `id` for each item and depend on that for expanding folders

# Screenshot

https://user-images.githubusercontent.com/105633534/225912321-7172f53f-dfdd-4a62-9876-2ef2d3d3586c.mov

### DIFF
--- a/plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/Folder.tsx
+++ b/plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/Folder.tsx
@@ -8,8 +8,8 @@ export type Props = {
   isMobile?: boolean;
   expandAll?: boolean;
   nestLevel: number;
-  expandedKeys: string[];
-  setExpandedKeys: React.Dispatch<React.SetStateAction<string[]>>;
+  expandedFolders: { id?: string; name?: string }[];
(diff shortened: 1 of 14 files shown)


plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/Folder.tsx
```diff
@@ -19,37 +19,40 @@ const Folder: React.FC<Props> = ({
...
-        if (prevState.includes(key)) {
-          const index = prevState.findIndex(item => item === key);
-          newExpandedKeys.splice(index, 1);
+    (folder: { id?: string; name?: string }) => {
+      setExpandedFolders((prevState: { id?: string; name?: string }[]) => {
+        const newExpandedFolders = [...prevState];
+        if (prevState.find(folder => (folder.id ? folder.id === id : folder.name === name))) {
+          const index = prevState.findIndex(folder =>
+            folder.id ? folder.id === id : folder.name === name,
+          );
```
KaWaite, line 40: I think you can simplify this by doing something like
```
 const index = prevState.findIndex(folder =>
            folder.id ? folder.id === id : folder.name === name,
          );
if (index){
   newExpandedFolders.splice(index, 1);
..
...
```

KaWaite (approved):

PR #296, opened 2023-03-23
refactor(plugin): merge infobox global property

airslice:
## Overview

This PR refactor some on merge infobox global property.

### DIFF
--- a/plugin/src/sidebar.ts
+++ b/plugin/src/sidebar.ts
@@ -653,7 +653,7 @@ function createLayer(dataset: DataCatalogItem, overrides?: any) {
             alpha: 0.8,
           },
         }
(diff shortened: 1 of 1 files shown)


plugin/src/sidebar.ts
```diff
@@ -643,37 +633,38 @@ function createLayer(dataset: DataCatalogItem, overrides?: any) {
...
+      ? {
+          raster: {
+            alpha: 0.8,
           },
-        },
-      },
-    },
-  );
+        }
+      : { ...(omit(overrides, "infobox") ?? {}) }),
```
KaWaite, line 656: this should be omit(overrides, ["data", "infobox"]) too right?

airslice: Updated.

KaWaite (approved):

PR #309, opened 2023-03-24
fix(plugin): some infobox attributes is missing (#309)Co-authored-by: rot1024 <aayhrot@gmail.com>

airslice:
## Overview

- Add `浸水ランク（独自分類）`
- Update name `構造種別（独自）`

## Not changed:

- Add `継続時間` -> It already there so didn't modify on this property.
- 土地利用（川崎市） → attrubtes is not shown -> This should be fixed yesterday night.
- 津波浸水想定区域（川崎） → infobox and attrubtes is not shown -> infobox & attributes are shown but the source data seems incorrect compare with v1.1


plugin/web/extensions/infobox/core/utils/attributes.ts
```diff
@@ -134,6 +134,7 @@ function floodFields(properties: Properties): any {
         const prefix = `${a["uro:description"]}（${a["uro:adminType"]}管理区間）_${a["uro:scale"]}`;
         return [
           [`${prefix}_浸水ランク`, a["uro:rank_code"]],
+          [`${prefix}_浸水ランク（独自分類）`, a["uro:rankOrg"]],
```
rot1024, line 137:
```suggestion
          [`${prefix}_浸水ランク`, a["uro:rank_code"] || a["uro:rankOrg"]],
```

rot1024 (changes requested):

plugin/web/extensions/infobox/core/utils/attributes.ts
airslice: Updated.

PR #350, opened 2023-03-27
fix(plugin): Can set Fill color(Gradient)(Point) component to group 

ghost:
# Overview
Can set Fill color(Gradient)(Point) component to group

# What's done
Group info is set in `value` property, which isn't added to `onUpdate` and is always overridden by `colorGradient`.
The solution is to override value properties but still add new value properties.
Fix `cleanseOverrides`

# Screenshots

https://user-images.githubusercontent.com/105633534/227899373-4c61f5d6-c2d5-4cfc-bcc3-f9de5b918062.mov

### DIFF
--- a/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/point/PointColorGradient.tsx
+++ b/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/point/PointColorGradient.tsx
@@ -97,6 +97,7 @@ const PointColorGradient: React.FC<BaseFieldProps<"pointColorGradient">> = ({
     if (isEqual(omit(colorGradient, "override"), omit(value, "override"))) return;
     const timer = setTimeout(() => {
       onUpdate({
+        ...value,
         ...colorGradient,
         override: {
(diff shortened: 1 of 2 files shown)


plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/Field/Fields/point/PointColorGradient.tsx
```diff
@@ -97,7 +97,13 @@ const PointColorGradient: React.FC<BaseFieldProps<"pointColorGradient">> = ({
...
     const timer = setTimeout(() => {
       onUpdate({
-        ...colorGradient,
+        ...value,
+        field: colorGradient.field,
+        min: colorGradient.min,
+        max: colorGradient.max,
+        step: colorGradient.step,
+        startColor: colorGradient.startColor,
+        endColor: colorGradient.endColor,
```
KaWaite, line 106: Can you not just do
```
onUpdate({
  ...value,
  ..colorGradient,
overide:{
...
})
```

KaWaite (approved):

PR #351, opened 2023-03-27
fix(plugin): 3d tiles appearance update

airslice:
## Overview

As the discuss result: considering setting timeline to 10.a.m. will cause more problem with datasets setting timeline, it seems better to go back to light option solution. So added the light option back with updated light params.

This PR:
- Add back light option.
- Set floodColor default tranparency to 1 as this should be controlled with transparency field component.
- Update floodColor for better matching the legend color.
- Set colorBlendMode to `highlight` with white color to keep texture when `water`.
- Set flood code level from 1 - 6. (it was 0 - 5 before so doesn't match the legend).

## What i didn't do

- For buildings, override color without colorBlendMode replace will make the color much more bright and match the legend color, but at the same time it will highlight with texture if change to LOD with textures, that looks not good so we didn't apply this.

## Test

Test project [here](https://test.reearth.dev/edit/01gwdyy5gwxa9nw8txksnbpn7z).

## Screen shot
WechatIMG25444
WechatIMG25445
WechatIMG25446
WechatIMG25447
WechatIMG25448


plugin/web/extensions/sidebar/core/components/desktop/hooks/projectHooks.ts
```diff
@@ -54,6 +54,14 @@ export const defaultProject: Project = {
       },
     ],
     atmosphere: { shadows: true },
+    light: {
```
KaWaite, line 57: This also needs to be updated in the mobile's projectHooks file too (sorry due to a lack of time I didn't consolidate shared hooks yet between desktop and mobile...)

airslice: Updated!

KaWaite (approved):

PR #312, opened 2023-03-24
feat(plugin): save data catalog's tab + support mobile

nourbalaha:
Done:
- Save data catalog's tab
- Save expandedFolders and searchTerm on mobile.

PR #386, opened 2023-03-31
fix(plugin): data catalog trees effect each other

airslice:
## Overview

There're two (or three) trees in data catalog, their item ids are like `node-${index}` so there're different items with the same id between trees. When user fold/unfold any folder, the items with the same id will also be considered as fold/unfold.

The most obvious example is the first item of the two trees. They both have id `node-0` and each time click on one of them, the other also got effected.

This PR adds an optinal prefix to `makeTree` and use `groupBy` as the prefix so that tree items has unique id now.


airslice:
Better hold on this one until [this](https://github.com/eukarya-inc/reearth-plateauview/pull/312) feature been merged.

airslice:
Since current UX is reset folders when switch tab. the fix can be skipped.

PR #391, opened 2023-03-31
fix(plugin): the legend icon doesn't take the right color

nina992:
# Overview
 The legend icon doesn't take the color that the user enters.

## What I've done
 - Use Canvas element to modify the image specified in the imageUrl prop with the specified blendColor, width, and height. The image modification process involves blending the original image with the specified blendColor to create a modified image. 
 - The `useModifiedImage` hook uses the` CanvasRenderingContext2D` API to create a new `ImageData` object for the modified image by iterating over every pixel in the original image, blending it with the specified `blendColor`, and writing the result to the new `ImageData `object. Once the modified image is complete, it is drawn onto the second canvas element and converted to a data URL for display.
 - The second canvas is important in this code because it is being used to apply a blend effect to the original image.. The blend effect is achieved by drawing the original image onto two different canvas elements.
The first canvas is used to draw the original image without any modification. This is done by calling the ctx.drawImage() method and passing the original image as the argument.
The second canvas is used to apply the blend effect to the original image. This is done by calling the ctx2.drawImage() method to draw the original image onto the canvas, and then setting the globalCompositeOperation property of the canvas context to "multiply". This causes any new shapes or images drawn onto the canvas to be blended using a multiplication blend mode.
After setting the blend mode, a filled rectangle is drawn on the canvas using the ctx2.fillRect() method, using the color value that was created from the blendColor prop. Finally, the globalCompositeOperation property is set to "destination-in", which means that any new shapes or images drawn on the canvas will only be visible where they overlap with the existing content. The original image is then drawn onto the canvas again using the ctx2.drawImage() method.
The modified image is created by calling canvas2.toDataURL() method, which returns a data URL representation of the canvas content. The modified image is then set as the modifiedImageUrl state variable and returned from the hook.
 
## Screenshot
image
image


plugin/web/sharedComponents/ModifiedImage/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (!ctx2) {
+        setLoading(false);
+        return;
+      }
+      ctx2.drawImage(image, 0, 0, width, height);
+      ctx2.globalCompositeOperation = "multiply";
+      ctx2.fillStyle = color;
+      ctx2.fillRect(0, 0, canvas2.width, canvas2.height);
+      ctx2.globalCompositeOperation = "destination-in";
+      ctx2.drawImage(image, 0, 0, width, height);
```
KaWaite, line 52: Why do you do this at the beginning and then do it again here? Is that on purpose?

plugin/web/sharedComponents/ModifiedImage/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+
+      const modifiedImage = new Image();
+      modifiedImage.src = canvas2.toDataURL();
+
+      setModifiedImageUrl(modifiedImage.src);
```
KaWaite, line 57: Is there any chance that modifiedImage.src will be undefined or any other issues? I feel like displaying the original image is good to fallback to

plugin/web/sharedComponents/ModifiedImage/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+      setLoading(false);
+    };
+    image.onerror = (event: string | Event) => {
+      if (
+        event instanceof Event &&
+        event.target instanceof HTMLImageElement &&
+        !event.target.complete
+      ) {
+        console.error("Failed to load image due to cross-origin error");
```
KaWaite, line 66: Are you actually able to assume that if there is an error, it'll 100% always be because of cross-origin? If so, this is fine. But if not, a more generic message is MUCH more helpful (or even no error log is more helpful than a wrong error log)

plugin/web/sharedComponents/ModifiedImage/hooks.ts
nina992: yes,The second canvas is important in this code because it is being used to apply a blend effect to the original image.. The blend effect is achieved by drawing the original image onto two different canvas elements.
The first canvas is used to draw the original image without any modification. This is done by calling the ctx.drawImage() method and passing the original image as the argument.
The second canvas is used to apply the blend effect to the original image. This is done by calling the ctx2.drawImage() method to draw the original image onto the canvas, and then setting the globalCompositeOperation property of the canvas context to "multiply". This causes any new shapes or images drawn onto the canvas to be blended using a multiplication blend mode.
After setting the blend mode, a filled rectangle is drawn on the canvas using the ctx2.fillRect() method, using the color value that was created from the blendColor prop. Finally, the globalCompositeOperation property is set to "destination-in", which means that any new shapes or images drawn on the canvas will only be visible where they overlap with the existing content. The original image is then drawn onto the canvas again using the ctx2.drawImage() method.
The modified image is created by calling canvas2.toDataURL() method, which returns a data URL representation of the canvas content. The modified image is then set as the modifiedImageUrl state variable and returned from the hook.

plugin/web/sharedComponents/ModifiedImage/hooks.ts
nina992: There is a chance that modifiedImage.src could be undefined in case of some issue while creating the modified image. For example, if the canvas element is not able to draw the image due to some error, the modifiedImage.src could be undefined.

nina992: Done

plugin/web/sharedComponents/ModifiedImage/hooks.ts
nina992: I added more generic error handling 
Done

plugin/web/sharedComponents/ModifiedImage/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+  const [modifiedImageUrl, setModifiedImageUrl] = useState("");
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    const image = new Image();
+    image.crossOrigin = "anonymous";
+    image.src = imageUrl;
+    image.onload = () => {
+      if (!image.complete) {
+        console.error("Failed to load image due to cross-origin error");
```
KaWaite, line 21: Another cross origin error that might be better more generic

KaWaite (approved):

plugin/web/sharedComponents/ModifiedImage/hooks.ts
nina992: done

PR #393, opened 2023-03-31
chore(plugin): refactor datasetHooks

pyshx:
In this refactor: we use the useEffect() hook to ensure that handleBackendFetch() is only called once the component has mounted, rather than during rendering. We also wrap the call to handleDataRequest() inside a try-catch block to catch and handle any errors that may occur. Additionally, we optimize the state updates triggered by handleBackendFetch() and handleDataRequest() to avoid unnecessary re-renders. Finally, we use async/await syntax for cleaner and more concise code.

### DIFF
--- a/plugin/web/extensions/sidebar/core/components/desktop/hooks/datasetHooks.ts
+++ b/plugin/web/extensions/sidebar/core/components/desktop/hooks/datasetHooks.ts
@@ -1,6 +1,6 @@
 import { Project } from "@web/extensions/sidebar/types";
 import { postMsg } from "@web/extensions/sidebar/utils";
-import { useCallback } from "react";
+import { useCallback, useEffect } from "react";
 
 import { Data, DataCatalogItem, Template } from "../../../types";
 import { convertToData } from "../../utils";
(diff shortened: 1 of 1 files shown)


plugin/web/extensions/sidebar/core/components/desktop/hooks/datasetHooks.ts
```diff
@@ -94,22 +109,30 @@ export default ({
   );
 
   const handleDatasetSave = useCallback(
-    (dataID: string) => {
-      (async () => {
-        if (!inEditor) return;
-        setLoading?.(true);
-        const selectedDataset = project?.datasets.find(d => d.dataID === dataID);
+    async (dataID: string) => {
```
KaWaite, line 112: Do we not also need to update the prop types where this and the `handleDatasetPublish` are being used?
I was kinda being lazy with types and trying to avoid that by doing the (async()=>{})() style :P
I guess we aren't returning anything in these callbacks so I guess the typings are fine. Is that correct?

plugin/web/extensions/sidebar/core/components/desktop/hooks/datasetHooks.ts
```diff
@@ -119,31 +142,36 @@ export default ({
...
-          method: "POST",
-          body: JSON.stringify({ id: dataset.itemId }),
-        })
-          .then(r => {
-            if (!r.ok)
-              throw `failed to publish the data on gspatial.jp: status code is ${r.statusText}`;
-          })
-          .catch(console.error);
+      try {
+        await handleDataRequest(dataset);
```
KaWaite, line 157: Any reason this handleDataRequest is inside the try/catch? Doesn't this mean if the handleDataRequest fails we will never do the geospatialjp fetch? That isn't good

PR #397, opened 2023-04-01
fix(plugin): type definition error on data catalog config data layer, support string layer

airslice:
## Overview

- Fix type definition error on data catalog config data item `.layer`.
- Support type string for layers.


rot1024 (approved):

airslice:
This is only the plan-B for[ this PR](https://github.com/eukarya-inc/reearth-plateauview/pull/396). SInce that one is okey, this PR will be closed.

PR #400, opened 2023-04-02
feat(plugin): show description of folders in datacatalog

ghost:
# Overview
This PR updates the handling of selectedItem to allow for the selection of either individual data catalog items or groups of items. The main changes include:

- Hide data with type `folder`
- Update the selectedItem type to `DataCatalogItem | DataCatalogGroup` in the `DatasetTree`, `DatasetsPage`, and `DataCatalog` components
- Modify the `handleSelect` callback function to handle the new selectedItem type.
- if a folder is selected get its appropriate dataset
- Remove the `handleOpenDetails` function and the onOpenDetails property, incorporating their functionality into handleSelect.
This update provides more flexibility in the selection process by allowing users to select both individual items and groups of items in the data catalog.

# Screenshot

https://user-images.githubusercontent.com/105633534/229347734-c3493f6c-a86a-4f4a-bd12-73c5f45e9521.mov


plugin/web/extensions/sidebar/popups/mobileDropdown/Catalog/index.tsx
```diff
@@ -95,9 +96,11 @@ const Catalog: React.FC<Props> = ({
   }, [backendURL, backendProjectName]);
 
   const handleOpenDetails = useCallback(
-    (data?: DataCatalogItem) => {
-      setSelectedDataset(data);
-      setPage("details");
+    (data?: DataCatalogItem | DataCatalogGroup) => {
+      if (data && "dataID" in data) {
```
KaWaite, line 100: Doesn't this mean no folders will be selectable with this callback?

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/index.tsx
```diff
@@ -62,10 +58,11 @@ const DatasetsPage: React.FC<Props> = ({
...
 
-  const selectedDataset = useMemo(
-    () => catalog?.find(item => item.dataID !== undefined && item.dataID === selectedDatasetID),
-    [catalog, selectedDatasetID],
-  );
+  const selectedDataset = useMemo(() => {
+    return selectedItem && "dataID" in selectedItem
+      ? catalog?.find(item => item.dataID === selectedItem.dataID)
+      : selectedItem;
+  }, [catalog, selectedItem]);
```
KaWaite, line 65: Sorry I know you explained this change, but can you explain why its needed still?
Also, I explained the need for checking if it is undefined.

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/Details.tsx
```diff
@@ -57,12 +57,12 @@ const DatasetDetails: React.FC<Props> = ({
       )}
     </>
   );
-  return dataset ? (
+  return dataset && ("dataID" in dataset || dataset.desc) ? (
     <DetailsComponent
       dataset={dataset}
-      addDisabled={addDisabled(dataset.dataID)}
+      addDisabled={"dataID" in dataset && addDisabled(dataset.dataID)}
```
KaWaite, line 63: Hm? Doesn't this mean that if there is no dataID it will never be disabled??

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/index.tsx
basel-i: @KaWaite , `selectedDataset` doesn't have all the properties needed for the details, so it will fetch the right data from the catalog (it always has been like this).
The edit I made is if the `selectedDataset` is not a `DataCatalogItem` get the `DataCatalogGroup ` itself as now it has a `desc` property.

as for undefined, my mistake I updated it now.

plugin/web/extensions/sidebar/modals/datacatalog/components/hooks.ts
```diff
@@ -37,7 +36,7 @@ export default () => {
   const processedCatalog = useMemo(() => {
     if (catalogData.length < 1 || data === undefined) return;
     const c = handleDataCatalogProcessing(catalogData, data);
-    return inEditor ? c : c.filter(c => !!c.public);
+    return inEditor ? c : c.filter(c => !!c.public || c.type_en === "folder");
```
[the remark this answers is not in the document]
basel-i: @KaWaite , actually the `makeTree` function gets the data from here, so it is needed.

plugin/web/extensions/sidebar/popups/mobileDropdown/Catalog/index.tsx
basel-i: @KaWaite , you are right, because on mobile we can't make folders selectable, it will open a full page, and make accessing files impossible.
So on mobile I made sure it keeps the old behavior.

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/Details.tsx
basel-i: @KaWaite , I'm hiding this section inside the `DetailsComponent ` so this parameter doesn't matter for items with no `dataID`.

plugin/web/extensions/sidebar/modals/datacatalog/components/hooks.ts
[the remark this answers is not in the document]
KaWaite: Ah sorry didn't realize that... understood

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/index.tsx
KaWaite: Now that I think about, since you consolidated `selectedDatasetID` and `selectedItem`, why do you think `selectedDataset` is even needed? Now we basically have CatalogItem1 already assigned to selectedItem, and then we are searching the catalog again for CatalogItem1 and assigning it to selectedDataset, and then using selectedDataset.
Can't we just use selectedItem as is?
If selectedItem isn't selected at the level where the folder and desc is already added I feel like our logic is wrong somewhere.

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/Details.tsx
KaWaite: Then you don't need this change, right? Its the same as in a conditional like this `if(dataset && dataset.components)` you don't need to do `dataset?.components` beacuse you have already ruled out an undefined dataset, right?

plugin/web/extensions/sidebar/popups/mobileDropdown/Catalog/index.tsx
KaWaite: But in the mobile's catalog we aren't even including folders, right? So we shouldn't need to change anything in the mobile code as the folders won't be there to confuse things, right?

basel-i: @KaWaite , you are right, but we changed the `onSelect` code, which affects both mobile and desktop, so `onSelect` can now have `DataCatalogGroup ` as a parameter, so I've added this condition for mobile here so it won't affect this part.

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/Details.tsx
basel-i: @KaWaite , actually this is needed, because `addDisabled ` function needs `dataID` to work, and `dataset` could be a `DataCatalogGroup ` so it will give a typing error even if you do something like you mentioned:
`Property 'dataID' does not exist on type 'DataCatalogItem | DataCatalogGroup'.`

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/index.tsx
basel-i: @KaWaite , it is needed, It is responsible to get the latest state of `DataCatalogItem ` each time, if we remove that part (useMemo) it will break the publishing feature here, it won't change it's status after each publish/unpublish.

plugin/web/extensions/sidebar/popups/mobileDropdown/Catalog/index.tsx
KaWaite: @basel-i Sure the onSelect can accept DataCatalogGroup, but this handleOpenDetails can and should only accept the DataCatalogItem. I think you would get a type error passing something like `dataset: Group | Item` in to something that only wants `Item`, but passing `dataset: Item` in to `dataset: Group | Item` shouldn't cause troubles. You know what I mean?

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/index.tsx
KaWaite: Then we should use the old selectedDatasetID rather than the selectedItem, right? That way we don't needlessly save the whole dataset object

plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/Details.tsx
KaWaite: Ah I see now. In the return's conditional we allow through even if dataset has no dataId, but has a desc. In that case, we won't have dataID. Gotcha

KaWaite (approved):

PR #413, opened 2023-04-07
fix(plugin): json property for usecase mvt data

airslice:
## Overview

We meet the use case that `usecase` data (mvt) also needs to apply style base on property value, therefore `jsonProperties: ["attributes"]` is needed.
- Note: on reearth side the json parse is wrapped with try catch so it should be okey if there's no attributes.

Also did some refactor or fix:

- Define `url` `layers` besides `format` and make the values out of `data` object.
  - Make the logic more clear.
  - The `format` is also been used in later codes, so previous code could make the data not match.
- Use `hasConfigData` as condition for all `format` / `url` / `layers`. 
  - They were handling independently before and once there's some issue on data (e.g. dataset.config?.data?.[0] has `url` but no `layer`, the dataset's layers will be used) the data could be not match.
- Remove unnecessary `tran` from the first condition.
  - `tran` could be mvt or 3d tiles. The json property should be added only when it's mvt and the second condition already did this.


airslice:
Ah i notice Piyush did refactor for `merge` in [this](https://github.com/eukarya-inc/reearth-plateauview/pull/412/files) PR. Maybe better merge that first and i'll update my PR then.

PR #414, opened 2023-04-10
fix(plugin): infinite loop on mobile

airslice:
## Overview

- Fix some filed component value can't be applied, or enter an apply-reset loop:
When on mobile, the `project` is not synced. Therefore the update of components can't get the updated values, when multiple filed components update triggerred, the out-of-date value could be carried and override the previous update.

- Fix infinite update:
switchDataset use `===` to check two objects, which could lead to infinite loop update.

- Fix overrides happen when initialization for `BuildingColor` `BuildingFilter` `BuildingShadow` `Clipping`.
- Fix override sort for `updateAt`, it could be string.
- Fix data catalog can't fetch data on switch tab.
- Fix an CSS property error.
- Fix mobile build search use wrong id.

## Test

Test project [here](https://test.reearth.dev/edit/01gxns5mw9axjt9q3sc2fajfnh).

### DIFF
--- a/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/hooks.ts
+++ b/plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/hooks.ts
@@ -7,11 +7,11 @@ import {
   postMsg,
 } from "@web/extensions/sidebar/utils";
 import { getActiveFieldIDs } from "@web/extensions/sidebar/utils/dataset";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useState, useRef } from "react";
 
 import { cleanseOverrides } from "../FieldComponent/fieldConstants";
 import generateFieldComponentsList from "../FieldComponent/fieldHooks";
-import { ConfigData } from "../FieldComponent/Fields/types";
+import { ConfigData, FieldComponent } from "../FieldComponent/Fields/types";
 
 export default ({
   dataset,
@@ -29,6 +29,7 @@ export default ({
   onOverride?: (dataID: string, activeIDs?: string[]) => void;
 }) => {
   const [activeComponentIDs, setActiveIDs] = useState<string[] | undefined>();
+  const latestComponents = useRef<FieldComponent[]>(dataset.components ?? []);
 
   useEffect(() => {
     const newActiveIDs = getActiveFieldIDs(
@@ -68,6 +69,7 @@ export default ({
       postMsg({ action: "unselect" });
       onDatasetUpdate?.({
         ...dataset,
+        components: latestComponents.current,
         selectedDataset,
       });
     },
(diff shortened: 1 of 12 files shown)


plugin/web/extensions/sidebar/core/components/content/common/FieldComponent/Fields/general/Template.tsx
```diff
@@ -22,6 +22,7 @@ const Template: React.FC<BaseFieldProps<"template">> = ({
   onSceneUpdate,
 }) => {
   const [fieldComponents, setFieldComponents] = useState<FieldComponentType[] | undefined>();
+  const latestFieldComponents = useRef<FieldComponentType[] | undefined>();
```
KaWaite, line 25: If you are adding this, do we still need the fieldComponents useState? Seems like they are going to be synced now.
I am curious though how this useRef helps/fixes anything. Can you explain here or on Slack how this change is helpful?

plugin/web/extensions/sidebar/utils/overrides.ts
```diff
@@ -74,11 +74,14 @@ export const mergeOverrides = (
 
   const needOrderComponents = components
     .filter(c => (c as any).userSettings?.updatedAt)
-    .sort(
-      (a, b) =>
-        ((a as any).userSettings?.updatedAt?.getTime?.() ?? 0) -
-        ((b as any).userSettings?.updatedAt?.getTime?.() ?? 0),
-    );
+    .sort((a, b) => {
```
KaWaite, line 77: Do you know why we are getting both Date and string values for updatedAt? I feel like we should be able to expect only one type for something like this.

plugin/web/extensions/sidebar/core/components/content/common/DatasetCard/hooks.ts
```diff
@@ -29,6 +29,7 @@ export default ({
   onOverride?: (dataID: string, activeIDs?: string[]) => void;
 }) => {
   const [activeComponentIDs, setActiveIDs] = useState<string[] | undefined>();
+  const latestComponents = useRef<FieldComponent[]>(dataset.components ?? []);
```
KaWaite, line 32: How is this different from using dataset.components directly?
I think everywhere you are using the latestComponents ref their is dataset in the callback/useEffect's dependencies. You also set latestComponents to dataset.components, so as far as I can tell it'll update in two places, no? 

For example `handleFieldAdd` is called, updates `latestsComponents.current`, then updates the dataset directly. Then, the dataset state gets updated, passes those changes through props back to this hooks file, and then re-updates the useRef. Am I misunderstanding something somewhere?

plugin/web/extensions/sidebar/utils/overrides.ts
airslice: It's from post message i think. On mobile all the value become string when doing the sort.

KaWaite: Instead of supporting both Date and String, can you just update where we are setting the updatedAt to be turned to a string so we can always expect the same thing?

PR #415, opened 2023-04-12
fix(plugin): mobile building search overrides

airslice:
This is base on [this](https://github.com/eukarya-inc/reearth-plateauview/pull/414) branch.

## Overview

Override system is built mostly base on dataset field components, and the building search override was implemented as a special field component to each dataset. Therefore it can't work if the dataset cards (or `Selection` panel) is not enabled.

On desktop it's almost okey since the building search is working with the Selection panel at most of time.
On mobile it's not working since there's only one popup, the building search & field components can't exist at the same time.

## What did i do

Since on mobile the field components & the building search will not exists at the same time, building search overrides can always take higher priority. So (only on mobile hooks):
- Store field components' overrides in `cachedDatasetOverrides`.
- When building search override, merge the override with field components' one.
- When building search close, override back with field componnts' one (merged with default 3dtiles show & color).

Also update the data structure for desktop building search override, so that it can be sorted correctly with other field components.


plugin/web/extensions/sidebar/core/components/mobile/hooks/index.ts
```diff
@@ -17,7 +17,7 @@ export default () => {
   const [reearthURL, setReearthURL] = useState<string>();
   const [backendURL, setBackendURL] = useState<string>();
   const [backendProjectName, setBackendProjectName] = useState<string>();
-  const [buildingSearch, setBuildingSearch] = useState<BuildingSearch>([]);
+  const [buildingSearch] = useState<BuildingSearch>([]);
```
KaWaite, line 20: What is going on here? Without the setter function this will forever just be an empty array...

plugin/web/extensions/sidebar/core/components/mobile/hooks/projectHooks.ts
```diff
@@ -204,6 +211,41 @@ export default ({
...
+        cachedDatasetOverrides.current.get(dataID) ?? {},
+      );
+
+      postMsg({
+        action: "updateDatasetInScene",
+        payload: { dataID, overrides },
+      });
+    },
+    [],
+  );
```
KaWaite, line 247: You aren't doing any sort of "cleanseOverride" kind of thing here. Are ALL override values that the building search changes values that will by default have values?

plugin/web/extensions/sidebar/core/components/mobile/hooks/index.ts
airslice: The `buildingSearch` state is used for pass into dataset card, from this PR it will not be used when on mobile. I left it here to avoid changing too much to components.
But it could be misleading. I'm removing them now.

plugin/web/extensions/sidebar/core/components/mobile/hooks/projectHooks.ts
airslice: buildingSearch will and always update `show` & `color`，this is the same as the cleanseOverride for `field` building search on desktop. 
Since on mobile building search overrides become independent and above dataset override systems, it can't join dataset cleanseOverride calculation, so the value here is used as a default value for 3d tiles buildings. 
Without this default value it will have some problem, for example building search filtered the buildings, and when building search close, the cached overrides from dataset does not contain `show`, and the `show` conditions will be left there.

KaWaite (approved):

PR #417, opened 2023-04-18
fix(plugin): wrap the long titles for folders in datacatalog modal

nina992:
# Overview
 When opening datacatalog folder and open dataset page ,if the folder name is very long we need to wrpe in multiple lines
## Screenshot
image
test project : https://test.reearth.dev/edit/01gya8qmmaj6qzj1qksg6nf20x


plugin/web/extensions/sidebar/modals/datacatalog/components/content/DatasetsPage/DatasetTree/Folder.tsx
```diff
@@ -88,7 +88,6 @@ const Wrapper = styled.div<{ isOpen?: boolean }>`
     isOpen
       ? "height: 100%;"
       : `
-  height: 29px;
```
KaWaite, line 91: This and the other height should be kept, but converted to min-height. Without this, the Folders that don't have long names become much shorter than before

KaWaite (approved):

PR #429, opened 2023-05-21
feat(plugin): support showing the text label based on feature( marker )’s properties

nina992:
# Overview 
- Now  in our plateau-side bar, the `use label` component only supprt make the all point(marker) same text label, which is not what we want
- We want to show the text label based on feature( marker )’s properties
## What I've done
- Modify the field component `use Label` and give the user the ability to add expression but it should start with `=` so the system will find the property name that the user wants to be in marker's label. 
- The use still has the ability to enter normal text to be label for all markers.
- The system accepts the expression to be `=property name` , `={property name}`, or `=${property name}`.
# Screenshots
image

# Memo 
The task : https://www.notion.so/eukarya/Development-Management-151ba655e06c4b499b0a8935b238c6dd?p=936f84442ff6417087b57d37cee18821&pm=s


plugin/web/extensions/sidebar/core/components/content/common/FieldComponent/Fields/point/PointLabel.tsx
```diff
@@ -75,6 +94,7 @@ const PointLabel: React.FC<BaseFieldProps<"pointLabel">> = ({ value, editMode, o
   useEffect(() => {
     if (isEqual(value, pointLabel)) return;
     const timer = setTimeout(() => {
+      console.log("generateConditions:", conditions);
```
KaWaite, line 97: Don't need

plugin/web/extensions/sidebar/core/components/content/common/FieldComponent/Fields/point/PointLabel.tsx
```diff
@@ -25,7 +31,20 @@ const PointLabel: React.FC<BaseFieldProps<"pointLabel">> = ({ value, editMode, o
...
+        let extractedValue = fieldValue;
+        if (fieldValue.startsWith("={") && fieldValue.endsWith("}")) {
+          extractedValue = fieldValue.substring(2, fieldValue.length - 1);
+        } else if (fieldValue.startsWith("=${") && fieldValue.endsWith("}")) {
+          extractedValue = fieldValue.substring(3, fieldValue.length - 1);
+        } else {
+          extractedValue = fieldValue.substring(1);
+        }
+        setConditions(generateConditions(extractedValue));
+      } else setConditions(undefined);
```
KaWaite, line 46: Similar to the `generateConditions` function, can you pull out the stripping of the field value in to a function of its own and return extractedValue.
Then the code here would be something like
```
const extractedValue = sanitize(fieldValue)
const conditions = generateConditions(extractedValue)
setConditions(conditions)
```
Also, put `generateConditions` and the new helper function (doesn't have to be named `sanitize`, there is probably a better name for it :P) outside of the react component. Doesn't need to be rerendered!

KaWaite (approved):

PR #430, opened 2023-05-23
feat(plugin): welcome screen new design for mobile

nina992:
# Overview
To remind user there may be some problem while using lod2 building with texture will lead the crash, we added a description on the first page of device  
image
image
image

# Links 
Design Link: https://www.figma.com/file/xWM8xrtIAh5oMj0seLd9Jd/view-2.0?type=design&node-id=3162-88058
test link :https://test.reearth.dev/published.html?alias=didifjhebb


plugin/web/sharedComponents/Icon/icons.ts
```diff
@@ -76,8 +78,7 @@ import { ReactComponent as GearWheel } from "./Icons/gearWheel.svg";
 import { ReactComponent as ExternalLink } from "./Icons/externalLink.svg";
 
 // Ant Design Icons
-import { HolderOutlined } from "@ant-design/icons";
-import { PlayCircleFilled } from "@ant-design/icons";
+import { HolderOutlined, PlayCircleFilled } from "@ant-design/icons";
 
 export { InboxOutlined, DownloadOutlined } from "@ant-design/icons";
```
pyshx, line 83: Can you import them all at once since they are all coming from @ant-design/icons?

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
```diff
@@ -185,13 +233,13 @@ const ButtonWrapper = styled.div<{ selected?: boolean }>`
...
+const CheckWrapper = styled.div<{ isMobile?: boolean }>`
   display: flex;
   flex-direction: row;
-  align-items: center;
-  align-self: center;
+  align-items: ${({ isMobile }) => (!isMobile ? "center" : "flex-start")};
+  align-self: ${({ isMobile }) => (!isMobile ? "center" : "flex-end")};
   gap: 8px;
-  margin-top: 50px;
+  margin-top: ${({ isMobile }) => (!isMobile ? "50px" : "8px")};
```
pyshx, line 242: Instead of `!isMobile` it is advised to check `isMobile` for consistency in the whole code, Don't you think?

plugin/web/sharedComponents/Icon/icons.ts
nina992: one of the import and the other is export so they should keep like this

plugin/web/extensions/sidebar/modals/welcomescreen/index.tsx
nina992: done

nina992: done

PR #432, opened 2023-05-26
feat(plugin): support custom appearance for local dataset

airslice:
## Overview

This PR supports customize project name / logo / theme color.

## How to

- Set custom logo & project name in sidebar's property.
- Set theme color in scene property -> publish theme (custom) -> select color. This will effect plateauview plugin and builtin plugin at the same time.

## Test
- Test project [here](https://test.reearth.dev/published.html?alias=hjheghjhcg).

## Screen shot

test reearth dev_published html_alias=hjheghjhcg


plugin/web/extensions/sidebar/core/components/content/Share/hooks.ts
```diff
@@ -135,15 +135,19 @@ function generatePrintView(payload?: string) {
   };
   const localizedDate = currentDate.toLocaleDateString("ja", options);
 
+  const themeColor = getComputedStyle(document.documentElement).getPropertyValue(
```
KaWaite, line 138: Can you use getComputedStyle inside a sandboxed iframe????

plugin/web/extensions/sidebar/core/components/mobile/index.tsx
```diff
@@ -91,9 +93,19 @@ const MobileSidebar: React.FC<Props> = ({ className }) => {
...
   return (
     <Wrapper className={className}>
-      <PlateauIcon icon="plateauLogo" size={114} wide />
+      {!useCustomProjectHeader && <PlateauIcon icon="plateauLogo" size={114} wide />}
+      {useCustomProjectHeader && (
+        <CustomProjectHeader>
+          {customLogo && <CustomLogo src={customLogo} />}
+          {customProjectName && <CustomProjectName>{customProjectName}</CustomProjectName>}
+        </CustomProjectHeader>
+      )}
```
KaWaite, line 108: Can you do a ternary operator instead? So fall-backing to the plateauLogo

plugin/web/extensions/sidebar/core/components/content/Share/hooks.ts
airslice: I think so. It does not require parent's window, right?

KaWaite: Oh my mistake I think

KaWaite (approved):

PR #457, opened 2023-07-31
feat(plugin): detect position and set default marker for csv from my data

airslice:
## Overview

When user using My Data to upload a local csv file or load from an online csv url:
- system should detect the lat/lng/height by col name.
- system should set a default marker appearance for csv items if valid position col been found.

Basically this feature will be used in published page.

## What i have done

- Find the position columes with hard coded keywords:
  - latitude: `"latitude", "lat", "緯度", "北緯"`
  - longitude: `"longitude", "lng", "lon", "経度", "東経"`
  - height: `"height", "altitude", "alt", "高度"`
- Add default marker appearnce if lat & lng exists.
- Add `additionalData` to dataset and set override when createLayer
- Modify the text on detail panel of Mydata.
- Fix fetching project sometimes fetch twice bug.

## Test

Test with [this](https://test.reearth.dev/edit/01gtk2bjcpy103bep25kj4a2ve) project.

### DIFF
--- a/plugin/web/extensions/sidebar/modals/datacatalog/components/content/YourDataPage/utils.ts
+++ b/plugin/web/extensions/sidebar/modals/datacatalog/components/content/YourDataPage/utils.ts
@@ -1,0 +1,28 @@
+export const getAdditionalData = (content: string | undefined, format: string) => {
+  if (!content) return undefined;
+  if (format === "csv") {
+    const header = content.split("\r\n")[0];
+    const cols = header.split(",");
+    const latColumn = cols.find(col =>
+      ["latitude", "lat", "緯度", "北緯"].includes(col.toLowerCase()),
+    );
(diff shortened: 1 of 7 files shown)


plugin/web/extensions/sidebar/modals/datacatalog/components/content/YourDataPage/utils.ts
```diff
@@ -0,0 +1,28 @@
+export const getAdditionalData = (content: string | undefined, format: string) => {
```
keiya01, line 1:
```suggestion
export const getAdditionalCSVData = (content: string | undefined, format: string) => {
```
I guess we should name it as CSV to avoid confusing.

keiya01 (approved):

pyshx (approved):

PR #473, opened 2023-09-29
fix(plugin): building filter conditions

airslice:


airslice:
After some mtg we decided to go [update range max](https://github.com/eukarya-inc/reearth-plateauview/pull/474) instead of this solution. So i'll close this PR.

